PR #8, opened 2016-07-30
Fix issue where child context was replacing instead of augmenting parent context

mikestead:
To ensure this behaves the same as `preact` I just brought over and updated the relevant tests from there. Happy to make any alterations.


developit:
Awesome! Thank you so much!!

PR #10, opened 2016-08-05
add support for render(obj) where obj has a toJSON method

matthewmueller:
This PR will allow you to pass in objects into render and if they have a `toJSON` property, it will call that first.

It allows you to do fancy things like this:

``` js
const App => (
  div.class('box')({ custom: true })
)

let src = render(App)
// <div class="box" custom></div>
```

Let me know if you're down for this feature or if there's anything you'd like me to change! If this works for you I'll start opening PRs for the other renderers.


developit:
Works for me! I don't think this interferes with the existing handling since VNodes have a pretty specific format that is unlikely to have a `.toJSON` method.

matthewmueller:
going to close this since I don't need this anymore :-)

PR #13, opened 2016-10-14
SVG XLink attributes are rendered incorrectly

wildlyinaccurate:
Hello! XLink attributes on SVG elements don't seem to be transformed. For example `<use xlinkHref="#foo">` is being rendered as-is, when Preact renders it (correctly) as `<use xlink:href="#foo">`.

I'm going to attempt to fix it myself, but in the meantime please enjoy this failing test case:

```
  1) render Basic JSX should render SVG elements:

      AssertionError: expected '<svg viewBox="0 0 360 360"><use xlinkHref="#foo"></use></svg>' to equal '<svg viewBox="0 0 360 360"><use xlink:href="#foo"></use></svg>'
      + expected - actual

      -<svg viewBox="0 0 360 360"><use xlinkHref="#foo"></use></svg>
      +<svg viewBox="0 0 360 360"><use xlink:href="#foo"></use></svg>

      at Context.<anonymous> (test/render.js:79:24)
```

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -172,7 +172,8 @@ export default function renderToString(vnode, context, opts, inner, isSvgMode) {
 		for (let i=0; i<len; i++) {
 			let child = children[i];
 			if (!falsey(child)) {
-				let ret = renderToString(child, context, opts, true, nodeName==='svg');
+				let childSvgMode = nodeName==='svg' ? true : nodeName==='foreignObject' ? false : isSvgMode,
+					ret = renderToString(child, context, opts, true, childSvgMode);
 				if (!hasLarge && pretty && isLargeString(ret)) hasLarge = true;
 				pieces.push(ret);
 			}
--- a/src/index.js
+++ b/src/index.js
@@ -49,7 +49,7 @@ let shallowRender = (vnode, context) => renderToString(vnode, context, SHALLOW);
 
 
 /** The default export is an alias of `render()`. */
-export default function renderToString(vnode, context, opts, inner) {
+export default function renderToString(vnode, context, opts, inner, isSvgMode) {
 	let { nodeName, attributes, children } = vnode || EMPTY,
 		isComponent = false;
 	context = context || {};
@@ -117,6 +117,9 @@ export default function renderToString(vnode, context, opts, inner) {
 				if (attributes['class']) continue;
 				name = 'class';
 			}
+			else if (isSvgMode && name.match(/^xlink\:?(.+)/)) {
+				name = name.toLowerCase().replace(/^xlink\:?(.+)/, 'xlink:$1')
+			}
 
 			if (name==='class' && v && typeof v==='object') {
 				v = hashToClassName(v);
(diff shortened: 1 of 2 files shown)


wildlyinaccurate:
I've got a working patch but I don't know if it's the best approach.

developit:
Will need to account for recursion inside of svgs, but this is a great start! Thanks a bunch, I'm back from vacation on Wednesday and will get this merged :)

SimenB:
Any news here?

I'm trying to use this because of developit/preact-compat#249 to get snapshottesting, but if the string differs from what's in the browser, it's hard to trust it

SimenB:
I don't think the patch included here is enough though, there's a lot more than just `href`.

https://facebook.github.io/react/blog/2015/10/07/react-v0.14.html#notable-enhancements

> React DOM now supports these SVG attributes, which render into namespaced attributes: xlinkActuate, xlinkArcrole, xlinkHref, xlinkRole, xlinkShow, xlinkTitle, xlinkType, xmlBase, xmlLang, xmlSpace.

developit:
It seems like this PR accounts for any `xlink:` prefix. I'm fine merging it.

developit:
If you grab the latest changes from master, the build+tests should pass now.

wildlyinaccurate:
Brilliant, thank you!

src/index.js
```diff
@@ -169,7 +172,7 @@ export default function renderToString(vnode, context, opts, inner) {
 		for (let i=0; i<len; i++) {
 			let child = children[i];
 			if (!falsey(child)) {
-				let ret = renderToString(child, context, opts, true);
+				let ret = renderToString(child, context, opts, true, nodeName==='svg');
```
developit: only thing I just noticed - I think we need to toggle into and out of SVG mode the same [way Preact's renderer does](https://github.com/developit/preact/blob/master/src/vdom/diff.js#L112):

```js
isSvgMode = nodeName==='svg' ? true : nodeName==='foreignObject' ? false : isSvgMode;
```

The reason for this is so that the String render switches into SVG mode when it enters into an `<svg>` element, but then switches back out of SVG mode if it leaves SVG via a `<foreignObject>` element.  Let me know if that makes sense :)

wildlyinaccurate: Oh, good catch!

I want to capture this in a test but because the `render` doesn't actually use `isSvgMode` except for handling the `xlink` attributes, the only way I can think of is to assert that `<svg><foreignObject xlinkFoo /></svg>` renders the attribute as `xlinkFoo` and not `xlink:foo`. This feels a bit gross to me; what do you think?

wildlyinaccurate: Also, would you prefer I changed `isSvgMode` to be a global flag like in preact/diff rather than an argument to `renderToString`?

developit: I actually think it's better off as an argument.

developit: as for the test - I'd do something like that but take it a little further and verify that it's active/inactive for descendants:

```js
// input:
<svg>
  <image xlinkHref="#foo" />
  <foreignObject>
    <div contentEditable />
  </foreignObject>
</svg>

// output:
`<svg>
  <image xlink:href="#foo" />
  <foreignObject>
    <div contentEditable />
  </foreignObject>
</svg>`
```

wildlyinaccurate:
So I ended up switching to a global `isSvgMode` to fix a bug where `isSvgMode===false` in nested SVG children (`<svg><g><image xlinkHref /></g></svg>`) but I've run into a brick wall now: I can't figure out how to set `isSvgMode` back to `true` after a foreignObject element.

wildlyinaccurate:
@developit forgot to ping you the other day to say that this can be reviewed again.

developit:
Released as [3.5.0](https://github.com/developit/preact-render-to-string/releases/tag/3.5.0)!

PR #14, opened 2016-11-05
handle arrays as class attributes

matthewmueller:
This PR allows arrays to be passed into class attributes and joined together. 

I think JSX does this for you, but if you're not using JSX it ends up turning arrays into `class={["foo", "baz"]}` to `class="0 1"`.

Unrelated, but there was another test that was failing on master: 
screenshot 2016-11-06 01 43 13. Wasn't exactly sure how to fix that.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -119,7 +119,7 @@ export default function renderToString(vnode, context, opts, inner) {
 			}
 
 			if (name==='class' && v && typeof v==='object') {
-				v = hashToClassName(v);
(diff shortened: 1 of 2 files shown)


developit:
Woah, totally forgot to comment on this - I'm actually toying with pulling complex `class` support out of Preact, since everyone just uses [classnames](https://npm.im/classnames).  It does deduping, empty removal, supports nested Objects within Arrays and nice stuff like that.

matthewmueller:
no worries! yah, that's sounds fine to me.

PR #23, opened 2017-01-23
Render empty value if render returns false

mikestead:
[React supports](https://facebook.github.io/react/docs/react-component.html#render) returning `null` or `false` from a render function to indicate nothing to render.

Currently returning false during a server render will render the string ‘false’. 

As we use inline conditional rendering quite heavily e.g. `return condition && <div>yo</div>`this is causing our server renders with `preact-compat` to surface a fair amount of these unwanted strings.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -58,7 +58,7 @@ export default function renderToString(vnode, context, opts, inner, isSvgMode) {
 	let pretty = opts.pretty,
 		indentChar = typeof pretty==='string' ? pretty : '\t';
 
-	if (vnode==null) {
+	if (vnode==null || vnode===false) {
 		return '';
 	}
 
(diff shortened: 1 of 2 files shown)


mikestead:
I guess this could be seen as a breaking change for people using preact directly? Not sure how likely that would be though.

developit:
Tech ically breaking yeah, no worries though. I think react also renders nothing for the value `true` - are you able to verify and add that if so?

mikestead:
I had a try but React will throw in dev mode if you return `true`.

```
render(): A valid React element (or null) must be returned. You may have returned undefined, an array or some other invalid object
```

developit:
Perfect, this looks good to me then.

PR #25, opened 2017-02-06
Close void tags with slashes to support XHTML

elliotdavies:
This fixes https://github.com/developit/preact-render-to-string/issues/24 by closing void element tags with a `/`. It updates relevant tests accordingly.


developit:
Silly question: do we care about adding a space before `/`? 😅

elliotdavies:
Hmm, I'd be surprised if it was ever an issue in practice but actually the specs (https://www.w3.org/TR/html5/syntax.html#void-elements, https://www.w3.org/TR/xhtml1/#guidelines) do call for it so probably best. I'll tweak that now.

developit:
Perfect, LGTM.

PR #31, opened 2017-03-30
When styles object is empty don’t render style attribute

mikestead:
Currently an empty styles object will server render the style attribute e.g. `<div style></div>` when what we really want is `<div></div>`. This is a slight inconsistency with React (and I think Preact?). 

This sounds quite benign but as we don't allow inline styles in our CSP rules the addition of this attribute is quite major.. especially as React Router's `Link` seems to default this to an empty object.

### DIFF
--- a/src/util.js
+++ b/src/util.js
@@ -46,7 +46,7 @@ export function styleObjToCss(s) {
 			str += ';';
 		}
 	}
-	return str;
+	return str || undefined;
 }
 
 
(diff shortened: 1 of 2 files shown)


developit:
Ahh very interesting!

developit:
Just need to double check a benchmark before I cut a release with this - it changes the return value of that util function so it can produce 2 types instead of a single type, but I think in this case it will be harmless.

PR #36, opened 2017-06-07
Fix jsx renderer double closing elements

### LINKED ISSUE #34: JSX renderer renders self closing tags wrong
Just noticed, that the jsx renderer double closes self closing tags like `meta` or `base`.

Test case:
```jsx
const render = require("preact-render-to-string/jsx");

const html = <html><head><base href="foo/" /></head></html>;
console.log(render(html));
// Logs:
// <html>
//   <head>
//     <base href="foo/" /></base>
//   </head>
// </html>
```

marvinhagemeister:
Fixes #34

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -197,7 +197,7 @@ export default function renderToString(vnode, context, opts, inner, isSvgMode) {
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


marvinhagemeister:
Also added a test case for that.

PR #37, opened 2017-06-12
Remove extra characters from style strings to match react.

gunn:


developit:
It seems like this could be an issue for diffing? Browsers normalize style.cssText to the format render-to-string was producing.

rschristian:
Going to close this on out due to age & lack of movement.

Thanks for the proposal though!

PR #46, opened 2017-09-08
Fix dangerouslySetInnerHTML in JSX mode (Fixes #42)

### LINKED ISSUE #42: dangerouslySetInnerHTML difference in renderToJsxString vs renderToString
**steps to reproduce**

use dangerouslySetInnerHTML with renderToJsxString

```javascript
> const h = require('preact').h
> console.log(require('preact-render-to-string')(<div dangerouslySetInnerHTML={{__html:'<span>foo</span>'}} />))
<div><span>foo</span></div>
undefined
> console.log(require('preact-render-to-string/jsx')(<div dangerouslySetInnerHTML={{__html:'<span>foo</span>'}} />))
<div
  dangerouslySetInnerHTML={
    Object {
      "__html": "<span>foo</span>"
    }
  }
>
</div>
undefined
```
**expected results**

dangerouslySetInnerHTML "executed" in either render method.

**actual results**

dangerouslySetInnerHTML "executed" only in renderToString, but not renderToJsxString.

**notes**

i'm using [preact-jsx-chai](https://github.com/developit/preact-jsx-chai) and would rather not use dangerouslySetInnerHTML for hardcoded html in my tests. i've tracked down the issue to this library, but feel free to tell me that it should be addressed in the other one. i didn't see a way to tell that library not to use the jsxString render function.

developit:
/cc @robthehall

### DIFF
--- a/src/jsx.js
+++ b/src/jsx.js
@@ -22,6 +22,9 @@ let prettyFormatOpts = {
 
 function attributeHook(name, value, context, opts, isComponent) {
 	let type = typeof value;
+	
+	// Use render-to-string's built-in handling for these properties
+	if (name==='dangerouslySetInnerHTML') return false;
 
 	// always skip null & undefined values, skip false DOM attributes, skip functions if told to
 	if (value==null || (type==='function' && !opts.functions)) return '';


robthehall:
this change works for me! thanks!

PR #49, opened 2017-09-21
Add support for mixed rendering

Saegrov:
Mixed rendering is a mode that works like shallow rendering, but where
you can specify a list of components you want to render even tho they
are deeply nested.

We have a use case where we are using shallow rendering and Cheerio for unit testing our Preact components. We are using a redux-connect like higher order component to connect some components to our global state. 
When testing a component, that uses another component that is wrapped in a HOC, the shallow renderer stops at the HOC. 

With mixed rendering we can white list components we want to render and test that the wrapped components are used correctly.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@yr/preact-render-to-string",
   "amdName": "preactRenderToString",
-  "version": "4.0.1",
+  "version": "4.0.2",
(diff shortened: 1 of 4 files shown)


Saegrov:
Had to do some commits to use this internally at our company. Closing this in favour of  https://github.com/developit/preact-render-to-string/pull/50

PR #50, opened 2017-09-27
Add support for mixed rendering

Saegrov:
Mixed rendering is a mode that works like shallow rendering, but where
you can specify a list of components you want to render even tho they
are deeply nested.

We have a use case where we are using shallow rendering and Cheerio for unit testing our Preact components. We are using a redux-connect like higher order component to connect some components to our global state.
When testing a component, that uses another component that is wrapped in a HOC, the shallow renderer stops at the HOC.

With mixed rendering we can white list components we want to render and test that the wrapped components are used correctly.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -248,6 +248,7 @@ function getFallbackComponentName(component) {
 	return name;
 }
 renderToString.shallowRender = shallowRender;
+renderToString.mixedRender = mixedRender;
 
 
 export {
--- a/src/index.js
+++ b/src/index.js
@@ -48,12 +48,27 @@ renderToString.render = renderToString;
 let shallowRender = (vnode, context) => renderToString(vnode, context, SHALLOW);
 
 
+/**
+ *  Only render elements, leaving Components inline as `<ComponentName ... />`
+ *  except those specified in fullyRenderedComponents.
+ *	This method is just a convenience alias for `render(vnode, context, { shallow:true, alwaysRenderedComponented: [] })`
+ *	@param {VNode} vnode	JSX VNode to render.
+ *	@param {Array} alwaysRenderedComponents		List of components that should be rendered with shallow rendering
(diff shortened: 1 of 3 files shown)


src/index.js
```diff
@@ -69,9 +84,12 @@ export default function renderToString(vnode, context, opts, inner, isSvgMode) {
 
 	// components
 	if (typeof nodeName==='function') {
+		let componentName = getComponentName(nodeName);
 		isComponent = true;
-		if (opts.shallow && (inner || opts.renderRootComponent===false)) {
-			nodeName = getComponentName(nodeName);
+		if (opts.shallow &&
+			(inner || opts.renderRootComponent===false) &&
+			!opts.alwaysRenderedComponents.includes(componentName)) {
```
developit: Needs to use `indexOf()!==-1` because this lib supports IE9+.

src/index.js
```diff
@@ -48,12 +48,27 @@ renderToString.render = renderToString;
...
 
+/**
+ *  Only render elements, leaving Components inline as `<ComponentName ... />`
+ *  except those specified in fullyRenderedComponents.
+ *	This method is just a convenience alias for `render(vnode, context, { shallow:true, alwaysRenderedComponented: [] })`
+ *	@param {VNode} vnode	JSX VNode to render.
+ *	@param {Array} alwaysRenderedComponents		List of components that should be rendered with shallow rendering
+ *	@param {Object} [context={}]	Optionally pass an initial context object through the render path.
+ */
+let mixedRender = (vnode, alwaysRenderedComponents = [], context) => {
```
developit: This file is size-constrained - would it be alright to add the `alwaysRenderedComponents` option, but not export another method (`mixedRender`)?  That would keep things reasonably minimal.

Saegrov: Sure! I'll update the PR.

Saegrov:
@developit I've updated the PR, but kept the unit tests in a separate file. Do you want me to move them as well?

PR #54, opened 2018-01-04
Add support for htmlFor

forabi:
We are aliasing `react` and `react-dom` to `preact-compat` and I noticed that our server produces markup that contains `htmlfor` instead of `for`, i.e. the `htmlFor` prop is passed as-is. This is breaking input labels.


sventschui:
Closing as this made its way to master in another PR a month ago

PR #55, opened 2018-02-05
Support de-camelCasing JSX html attributes

blakef:
Taken the list of [Preact supported attributes](https://github.com/developit/preact/blob/f717ee8e53248d75ff8c1a46966adeec295d339b/src/preact.d.ts#L547) and transform these attributes when building the rendered string.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -24,6 +24,50 @@ const VOID_ELEMENTS = [
 	'wbr'
 ];
 
+const MAPPED_ATTRIBUTES = {
+	acceptCharset:     'accept-charset',
+	accessKey:         'access-key',
+	allowFullScreen:   'allow-full-screen',
+	allowTransparency: 'allow-transparency',
+	autoPlay:          'auto-play',
+	cellPadding:       'cell-padding',
+	cellSpacing:       'cell-spacing',
+	charSet:           'char-set',
+	className:         'class-name',
+	colSpan:           'col-span',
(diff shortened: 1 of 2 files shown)


tconroy:
We just ran into this bug today regarding srcSet attribute, would love to see this guy merged in :)

src/index.js
```diff
@@ -24,6 +24,50 @@ const VOID_ELEMENTS = [
...
+	mediaGroup:        'media-group',
+	minLength:         'min-length',
+	noValidate:        'no-validate',
+	radioGroup:        'radio-group',
+	readOnly:          'read-only',
+	rowSpan:           'row-span',
+	spellCheck:        'spell-check',
+	srcDoc:            'src-doc',
+	srcLang:           'src-lang',
+	srcSet:            'src-set',
```
gpoitch: `'srcset'`

src/index.js
```diff
@@ -24,6 +24,50 @@ const VOID_ELEMENTS = [
...
+	tabIndex:          'tab-index',
```
gpoitch: `'tabindex'`

sventschui:
@developit the RegEx approach would suffer from the issue of transforming `tabIndex` to `tab-index` instead of `tabindex`.

Combining both with a list of html attributes that need the hyphen to be removed might be a way to go.

Any other ideas?

PR #60, opened 2018-08-07
[feature] replace `undefined` prop values with defaultProps

ftes:
- mimick preact behaviour
- fix developit/preact-render-to-string#58

### DIFF
--- a/src/util.js
+++ b/src/util.js
@@ -39,15 +39,37 @@ export function styleObjToCss(s) {
 	return str || undefined;
 }
 
+/**
+ * Copy all properties from `props` onto `obj`.
+ * @param {object} obj Object onto which properties should be copied.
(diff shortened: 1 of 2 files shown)


src/util.js
```diff
@@ -39,15 +39,37 @@ export function styleObjToCss(s) {
...
+ * @param {object} props Object from which to copy properties.
+ * @returns {object}
+ * @private
+ */
 export function assign(obj, props) {
 	for (let i in props) obj[i] = props[i];
 	return obj;
 }
 
+/**
```
ftes: Copied from [preact vdom](https://github.com/developit/preact/blob/master/src/vdom/index.js#L47).

developit: A+!

PR #66, opened 2018-11-02
Adds support for `componentDidCatch()`

johnhaitas:
This does not include support for the `info` argument because `preact`'s support for `componentDidCatch()` does not pass this argument (https://github.com/developit/preact/pull/819#issue-136629967)

Also includes 3 new tests.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -73,9 +73,15 @@ function renderToString(vnode, context, opts, inner, isSvgMode) {
 				c._disable = c.__x = true;
 				c.props = props;
 				c.context = context;
-				if (nodeName.getDerivedStateFromProps) c.state = assign(assign({}, c.state), nodeName.getDerivedStateFromProps(c.props, c.state));
-				else if (c.componentWillMount) c.componentWillMount();
-				rendered = c.render(c.props, c.state, c.context);
+				try {
+					if (nodeName.getDerivedStateFromProps) c.state = assign(assign({}, c.state), nodeName.getDerivedStateFromProps(c.props, c.state));
+					else if (c.componentWillMount) c.componentWillMount();
+					rendered = c.render(c.props, c.state, c.context);
+				}
+				catch (error) {
+					if (c.componentDidCatch) c.componentDidCatch(error);
+					else throw error;
+				}
 
 				if (c.getChildContext) {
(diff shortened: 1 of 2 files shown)


johnhaitas:
This should resolve https://github.com/developit/preact-render-to-string/issues/64

johnhaitas:
This current implementation is incorrect per this documentation: https://reactjs.org/docs/react-component.html#error-boundaries
> **NOTE**
> Error boundaries only catch errors in the components *below* them in the tree. An error boundary can’t catch an error within itself.

Updating the code, starting with the tests.

johnhaitas:
I have updated the tests to verify that the error is in fact being passed to `componentDidCatch` with c7aba79.

src/index.js
```diff
@@ -82,7 +83,17 @@ function renderToString(vnode, context, opts, inner, isSvgMode) {
 				}
 			}
 
-			return renderToString(rendered, context, opts, opts.shallowHighOrder!==false);
```
developit: Idea: we can circumvent having to catch and re-throw the error when there's no componentDidCatch:

```js
if (c && c.componentDidCatch) {
	try {
		return renderToString(rendered, context, opts, opts.shallowHighOrder!==false);
	}
	catch (error) {
		c.componentDidCatch(error);
		rendered = c.render(c.props, c.state, c.context);
		// now we fall through to the non-componentDidCatch render!
	}
}
return renderToString(rendered, context, opts, opts.shallowHighOrder!==false);
```

developit:
Just noticed the React docs seem to indicate they don't support componentDidCatch during SSR:
https://reactjs.org/docs/error-boundaries.html

Are we early on this and risking breaking compat?

developit:
One thing I think we might want to consider is extending this to also allow catching enqueued state updates. Something like "render the component, then check if it errored or was dirtied via setState()/forceUpdate(). If so, render again".

johnhaitas:
per react's server support for error boundaries (https://github.com/developit/preact-render-to-string/pull/66/#issuecomment-464513856) ... if react does not support it, then it probably is not right to support it

marvinhagemeister:
I concur, this would be an awesome feature to support imo 👍

rschristian:
Closing as implemented in #305, upon the work done here.

Thanks!

PR #67, opened 2018-11-02
 Adds support for `static getDerivedStateFromError(error)`

johnhaitas:
This should resolve https://github.com/developit/preact-render-to-string/issues/65.

It includes code from https://github.com/developit/preact-render-to-string/pull/66. It is kept as a separate PR deliberately because `preact` doesn't currently have incoming support for `static getDerivedStateFromError(error)` https://github.com/developit/preact/pull/819

Documentation here: 
https://reactjs.org/docs/react-component.html#static-getderivedstatefromerror


johnhaitas:
This current implementation is incorrect per this documentation: https://reactjs.org/docs/react-component.html#error-boundaries
> **NOTE**
> Error boundaries only catch errors in the components *below* them in the tree. An error boundary can’t catch an error within itself.

Updating the code, starting with the tests.

src/index.js
```diff
@@ -73,9 +73,16 @@ function renderToString(vnode, context, opts, inner, isSvgMode) {
 				c._disable = c.__x = true;
 				c.props = props;
 				c.context = context;
-				if (nodeName.getDerivedStateFromProps) c.state = assign(assign({}, c.state), nodeName.getDerivedStateFromProps(c.props, c.state));
-				else if (c.componentWillMount) c.componentWillMount();
-				rendered = c.render(c.props, c.state, c.context);
+				try {
+					if (nodeName.getDerivedStateFromProps) c.state = assign(assign({}, c.state), nodeName.getDerivedStateFromProps(c.props, c.state));
+					else if (c.componentWillMount) c.componentWillMount();
+					rendered = c.render(c.props, c.state, c.context);
```
rpetrich: You likely will want to wrap the recursive call to `renderString` in try/catch instead of the call to the component's `render` method so only errors of children will be caught.

johnhaitas: yea, this was a correct observation, it has been fixed

johnhaitas:
I have updated the tests to verify that the error is in fact being passed to `getDerivedStateFromError` with 03971f9 or `componentDidCatch` with c7aba79.

developit:
I commented on #66 but I think it's slightly different here - it seems like React has indicated `getDerivedStateFromError()` is not going to be supported in string renderer (ever?). Not sure what we should do.

johnhaitas:
i'd say this is not a good idea if react does not support it

rschristian:
Closing as implemented in #305, upon the work done here.

Thanks!

PR #71, opened 2019-03-07
Get all tests passing

sync:
I understand the preact-x branch is just a draft but this PR fixes all the failing tests.

The only fixes that are looking suspicious are those one: `match.falsy`. The state appear to be null sometimes

### DIFF
--- a/test/render.js
+++ b/test/render.js
@@ -205,9 +205,7 @@ describe('render', () => {
 				.and.calledWithExactly(
 					match({
 						foo: 1,
-						children: [
-							match({ nodeName: 'span', children: 'asdf' })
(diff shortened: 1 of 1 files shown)


test/render.js
```diff
@@ -254,8 +252,8 @@ describe('render', () => {
 				.to.have.been.calledOnce
 				.and.calledWithExactly(
 					match(PROPS),
-					match({}),	// empty state
```
sync: with preact 10,  state can now null ?

marvinhagemeister: yup, we aligned our behaviour with `react`'s which initializes it with `null`.

PR #77, opened 2019-03-15
Resolve crash when using hooks

vaneenige:
This PR changes the external option for the build script from `none` to `preact` which resolves the crash when using hooks. (https://github.com/developit/preact/issues/1373)

The test for hooks ran fine when using the `src`, but crashed when using `dist` (the __h of undefined error). The reason for this was that with external option set to `none` resulted in the imported options from preact being an empty object (so `options.render` would not be called and hooks would crash). 

https://github.com/developit/preact-render-to-string/blob/8536c758e3d1808bcc18d64cd944da609531d11e/src/index.js#L66

Result difference in build:

**Before:** 
       `1510 B (mjs)`
       `1567 B (js)`

**After:**
	`1487 B (mjs)`
        `1549 B (js)`

**Difference:**
	`-23 B (mjs)`
	`-18 B (js)`

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
   "jsnext:main": "dist/index.mjs",
   "scripts": {
     "build": "npm run -s transpile && npm run -s transpile:jsx && npm run -s copy-typescript-definition",
-    "transpile": "echo 'export const ENABLE_PRETTY = false;'>env.js && microbundle src/index.js -f es,umd --target web --external none",
+    "transpile": "echo 'export const ENABLE_PRETTY = false;'>env.js && microbundle src/index.js -f es,umd --target web --external preact",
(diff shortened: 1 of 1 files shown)


sync:
Tested this and with this fix no more crash, thank you

didymu5:
Thanks for responding. hmm... here's what I have now. Same result.
```
"devDependencies": {
    "eslint": "^4.5.0",
    "identity-obj-proxy": "^3.0.0",
    "if-env": "^1.0.0",
    "jest": "^21.2.1",
    "microbundle": "^0.6.0",
    "node-sass": "^4.5.3",
    "preact-cli": "^2.0.2",
    "preact-widget-scripts": "^0.6.0",
    "sass-loader": "^6.0.6"
  },
  "dependencies": {
    "preact": "^10.0.0-rc.1",
    "preact-compat": "^3.17.0",
    "preact-habitat": "^3.1.0",
    "preact-render-to-string": "^5.0.6",
    "preact-router": "^2.5.7"
  },
```

PR #79, opened 2019-03-19
Add support for Fragments

### LINKED ISSUE #78: Fragments aren't rendered
`render` quietly does nothing when it encounters a `Fragment`.

[Here's a demo](https://codesandbox.io/s/lxrrv6r32z) showing the issue.

marvinhagemeister:
Fixes #78 :tada:


marvinhagemeister:
Good idea! Added a test for sibling `Fragments` for the pretty renderer suite :+1: Yeah I know what you mean. We spent so much time on getting the client side implementation right. The SSR part is ridiculously easy in comparison :joy:

PR #81, opened 2019-03-24
Add note about rendering back in the browser

codeclown:
I was confused by Preact appending to the container, and it took a while to find the answer in https://github.com/developit/preact-render-to-string/issues/11. I propose adding a note about it in the readme.

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,6 +93,24 @@ app.get('/:fox', (req, res) => {
 ---
 
 
+### Re-render Preact in the browser
+
+Note that unlike React, Preact does not overwrite pre-rendered DOM unless instructed to. Specify the third argument for `render` to match behaviour.
+
+```js
+// Previously rendered to <div id="root"> in the server
+var root = document.getElementById('root');
+render((
+	<Box type="open">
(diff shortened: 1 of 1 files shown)


marvinhagemeister:
Note that this behaviour is no more in Preact X. There we share the same behaviour add react and don't have a third argument.

codeclown:
Is Preact X the official name of a future Preact version? In that case it could be elaborated in this PR somehow. Would you suggest phrasing it e.g. "Note about Preact versions 9 and lower: unlike React, Preact does not overwrite pre-rendered DOM unless instructed to. Specify the third argument for `render` to match behaviour.".

marvinhagemeister:
Preact X (=10.0.0) is our next major version that is currently in alpha and should be released soon. It includes a few (mostly internal) breaking changes and we were originally aiming to do a v9 release which would just include some `vnode` changes. But it looks like that won't happen anymore. The release announcement has [more information](https://github.com/developit/preact/releases/tag/10.0.0-alpha.0) about Preact X :+1:

PR #83, opened 2019-03-31
Added type definition tests

9renpoto:
- Fixed type definitions

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -1,0 +1,27 @@
+{
+    "compilerOptions": {
+        "module": "commonjs",
+        "lib": [
+            "es6",
+            "dom"
+        ],
+        "noImplicitAny": true,
+        "noImplicitThis": true,
+        "strictNullChecks": true,
+        "strictFunctionTypes": true,
+        "baseUrl": "../",
+        "typeRoots": [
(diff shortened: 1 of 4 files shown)


tsconfig.json
```diff
@@ -0,0 +1,27 @@
+{
+    "compilerOptions": {
+        "module": "commonjs",
+        "lib": [
+            "es6",
+            "dom"
+        ],
+        "noImplicitAny": true,
```
andrewiggins: Could we simplify this line and the next couple to `strict: true`?

tsconfig.json
```diff
@@ -0,0 +1,27 @@
...
+        "noImplicitThis": true,
+        "strictNullChecks": true,
+        "strictFunctionTypes": true,
+        "baseUrl": "../",
+        "typeRoots": [
+            "../"
+        ],
+        "jsx": "react",
+        "jsxFactory": "h",
+        "experimentalDecorators": true,
```
andrewiggins: Let's remove this line since this library doesn't use decorators

9renpoto:
Thank you for the review. fixed it :)

9renpoto:
rebased

PR #88, opened 2019-05-01
Support createContext of preact x

### LINKED ISSUE #73: createContext Consumer does not receive context
Rendering components using createContext fails. 
Consumer components does not receive the context set in the Provider component. 

`renderToString` should check if there's a contextType and a corresponding context.

sventschui:
Adds support for createContext introduced by preact x.

I'm not sure whether these checks should also be introduced for class components. As the context consumer is currently a functional component it is sufficient to add the checks for these.

Fixes #73

### DIFF
--- a/test/context.js
+++ b/test/context.js
@@ -1,0 +1,33 @@
+import render from '../src/jsx';
+import { h, createContext } from 'preact';
+import chai, { expect } from 'chai';
+import sinonChai from 'sinon-chai';
+chai.use(sinonChai);
+
+// tag to remove leading whitespace from tagged template literal
+function dedent([str]) {
+	return str.split( '\n'+str.match(/^\n*(\s+)/)[1] ).join('\n').replace(/(^\n+|\n+\s*$)/g, '');
+}
+
+describe('context', () => {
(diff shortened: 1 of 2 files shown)


test/context.js
```diff
@@ -0,0 +1,52 @@
...
+				</Consumer>
+			</Provider>
+		);
+
+		expect(rendered).to.equal(dedent`
+			<section>value is: correct</section>
+		`);
+	});
+
+	it('should support falsy context value', () => {
```
marvinhagemeister: Would love to see another test for nested `Provider` components of the same `Context`. In that case always the nearest Provider value should be used. The implementation in this PR very likely already supports this, just a test is missing :slightly_smiling_face: 

```jsx
const { Provider, Consumer } = createContext();
<Provider value="foo">
  <Provider value="bar">
    <Consumer>
      {value => value === "bar"} // must be bar
    </Consumer>
  </Provider>
</Provider>
```

sventschui:
Actually the implementation is copied from the core ;)

Added the test for the nested `Provider`

PR #92, opened 2019-05-10
Suspense

sventschui:
**EDITED**

After a first approach that did a complete re-render for each suspension I pushed now a complete new version that incorporates the asynchronous-ness into `renderToString` directly. It will do the complete rendering in one go.

This PR adds a `allowAsync` option to `renderToString`s `options` argument. It indicates that `renderToString` is **allowed** to return a promise. This is the case whenever a suspension is found. (when this is false and a suspension is found it (the Promise) is thrown upwards as before).

Suspensions are processed in parallel. This means that rendering is not paused until a suspension is finished but other components/children are rendered.

I also introduced a `prepass.js` file making it possible to module-alias `react-ssr-prepass` to `preact-render-to-string/prepass` and thus enable compatibility with next.js canary.

What we might want to address is that the prepass replacement in `prepass.js`/`src/prepass.js` will result in a duplicate render. We might want to store the result of the prepass somewhere and return it when render is called on the same vnode. In this case we would need to validate that there are no different options used for the direct renderToString call than when calling it from prepass as the cached result will be wrong then. 

**NOTE:** Tests are currently failing as `lazy` is not yet part of a published version of `preact`. They will work if you link `preact#master`.

**Note2:** By accident this implementation also allows to return Promises from render/function components. I'd rather suppose we will warn/throw in this case, as this will not work in preact in the browser.


test/render.js
```diff
@@ -778,4 +778,34 @@ describe('render', () => {
...
+				<section>
+					<Suspense fallback={<div>Fallback...</div>}>
+						<article>
+							<Lazied />
+						</article>
+					</Suspense>
+				</section>
+			);
+
+			renderAsync(toBeRendered)
```
marvinhagemeister: The `Promise` needs to be returned, otherwise the test will be marked as complete before the promise is resolved

sventschui: Good catch. Thx!

test/render.js
```diff
@@ -253,7 +253,7 @@ describe('render', () => {
 				.to.have.been.calledOnce
 				.and.calledWithExactly(
 					match(PROPS),
-					match.falsy,
+					match({}),
```
sventschui: These changes were needed for the golfed away state initialisation in suspense

oscartbeaumont:
Is progress still being made to merge this as it would be very helpful for a project I am working on?

sventschui:
As @developit mentioned in Slack we should wait with these kind of things or move it to a non-official package as React itself has not yet any opinion about this topic. In the meantime you can use https://github.com/sventschui/preact-ssr-prepass.

oscartbeaumont:
That makes sense but I tried using the `preact-ssr-prepass` package and could not get it to work, I will open an issue.
Also, what does this mean for preactjs/preact#1756 which has this PR on its checklist?

sventschui:
IMHO we should close this PR and thus remove it from preactjs/preact#1756

Happy to look into the issue you face once you reported it :)

marvinhagemeister:
@sventschui alright, removed it from the list :slightly_smiling_face:

PR #93, opened 2019-05-11
Introduce renderToNodeStream

sventschui:
This PR aims to introduce `renderToNodeStream`.

`renderToNodeStream` is especially useful when used together with `Suspense`. This will allow SSR to send HTML to the client as soon as possible.

TODO:
- [x] Ensure compat with `ReactDOMServer.renderToNodeStream`
- [x] Add some more tests
- [ ] Wait for the next release of microbundle since the current one does not allow generators
- [x] Find a solution for backpressure (see comments)


developit:
Hmm - it seems like in order to support backpressure, we'd need to render directly into the stream rather than rendering to a String and converting after. Thoughts?

sventschui:
I'm not 100% familiar with backpressure in node. From what I understand we would need to create a readable stream that pauses rendering until `_read` is called (again). 

I'll definitely look into this when I find some time and if nobody does so before I do :)

package.json
```diff
@@ -40,16 +40,23 @@
   },
   "babel": {
     "presets": [
-      "env"
+      [
+        "@babel/env",
+        {
+          "exclude": [
```
ForsakenHarmony: should probably set the target to the minimum node version here

sventschui: Well `renderToNodeStream` requires node 10. The other parts of this package work fine with 9 and more importantly in the browser

ForsakenHarmony: just a bit concerned about the async generator

sventschui: I get that point. The `renderToNodeStream` should only be used in node anyways, thus an async generator feels good to me :)

I'd propose to disallow async generators in eslint as bundling regenerator would result in a massive bundle bloat. That will make sure we don't use it in modules targeted to browsers

developit:
This is great and the backpressure support is on point.

One question: does the async support here work with the Suspense API? It seems like it's handling the throwing of a Promise, but it would be doing so quite differently than if there was a <Suspense> root in place.  Maybe we should hold off on that bit?

README.md
```diff
@@ -90,6 +92,22 @@ app.get('/:fox', (req, res) => {
 ```
 
 
+### Render JSX / Preact / Whatever to a Node.js stream
+
+```js
+import { renderToNodeStream } from 'preact-render-to-string';
```
developit:
```suggestion
import { renderToNodeStream } from 'preact-render-to-string/stream';
```

sventschui:
Can you elaborate a bit on the concerns of the "`Suspense`" behaviour? The current implementation aims to have the same behaviour as if one would use `react-ssr-prepass` (not falling back to Suspense' fallback but awaiting all thrown promises)

One thing that is currently not too nice is that the rendering is blocked until one promise resolves but I can't come up with a simple solution to that one.

src/stream.js
```diff
@@ -0,0 +1,264 @@
...
+import { options, Fragment } from 'preact';
+import stream from 'stream';
+
+// components without names, kept as a hash for later comparison to return consistent UnnamedComponentXX names.
+const UNNAMED = [];
+
+const VOID_ELEMENTS = /^(area|base|br|col|embed|hr|img|input|link|meta|param|source|track|wbr)$/;
+
+
+function PreactReadableStream(vnode, context, opts) {
```
marvinhagemeister: I think it's safe to use a class here. The current `node` LTS version supports them natively.

src/stream.js
```diff
@@ -0,0 +1,264 @@
...
+			if (options.render) options.render(vnode);
+
+			let renderedVNode;
+
+			if (!nodeName.prototype || typeof nodeName.prototype.render!=='function') {
+				// Necessary for createContext api. Setting this property will pass
+				// the context value as `this.context` just for this component.
+				let cxType = nodeName.contextType;
+				let provider = cxType && context[cxType.__c];
+				let cctx = cxType != null ? (provider ? provider.props.value : cxType._defaultValue) : context;
```
loklaan: Hello! Looks like the use of `_defaultValue` might be a bug (#130); changing it to `__` should fix it.

Doc999tor:
Hey, guys
Any thoughts will this PR will be merged soon?
Can I offer some help?

ilanbm:
Any progress here?

src/stream.js
```diff
@@ -0,0 +1,264 @@
+import { encodeEntities, styleObjToCss, assign, getChildren } from './util';
+import { options, Fragment } from 'preact';
+import stream from 'stream';
+
+// components without names, kept as a hash for later comparison to return consistent UnnamedComponentXX names.
+const UNNAMED = [];
```
sventschui: Note to self: Seems like a memory leak when keeping references to components around in a global variable

muturgan:
everybody already wants to use it in production! how long?

JoviDeCroock:
Superseded by #296

PR #102, opened 2019-07-10
Join adjacent text nodes in pretty mode

marvinhagemeister:
This PR joins adjacent text nodes together instead of putting them on a new line.

```jsx
// source
<div>hello{' '} <b /></div>

// before
<div>
  hello
  {' '}
   
  <b />
</div>

// after
<div>
  hello{' '} 
  <b />
</div>
```

PR #103, opened 2019-07-10
Collapse adjacent text

developit:
Another approach to #102.

### DIFF
--- a/test/jsx.js
+++ b/test/jsx.js
@@ -30,6 +30,61 @@ describe('jsx', () => {
 		`);
 	});
 
+	it('should collapse whitespace', () => {
+		expect(renderJsx(
+			<p>a<a>b</a></p>
+		)).to.equal(dedent`
+			<p>
+				a
+				<a>b</a>
+			</p>
+		`);
+
+		expect(renderJsx(
+			<p>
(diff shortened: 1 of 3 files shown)


test/jsx.js
```diff
@@ -30,6 +30,61 @@ describe('jsx', () => {
...
+		)).to.equal(dedent`
+			<p>
+				a
+				<a>b</a>
+			</p>
+		`);
+
+		expect(renderJsx(
+			<p>
+				a{''}
```
marvinhagemeister: Idea for another test case:

```jsx
<p>
 {' '}a
 <a>b</a>
</p>
```

test/jsx.js
```diff
@@ -30,6 +30,61 @@ describe('jsx', () => {
...
+		`);
+
+		expect(renderJsx(
+			<p>
+				a{' '}
+				<a>b</a>
+			</p>
+		)).to.equal(dedent`
+			<p>
+				a
```
marvinhagemeister: I'm not sure if the expected result is correct. It's missing the trailing space from the JSX version.

marvinhagemeister:
Superseeded by #105

PR #122, opened 2019-10-15
Fix missing context with contextType for class components

### LINKED ISSUE #121: renderToString issue with context
There is an issue with class components and `renderToString`. Reproducible code:

```jsx
import React, { Component } from 'react';
import render from 'preact-render-to-string';

const Ctx = React.createContext();

class ClassConsumer extends Component {
    render() {
        const value = this.context;
        return <div>{value}</div>;
    }
}
ClassConsumer.contextType = Ctx;

const FuncConsumer = () => <Ctx.Consumer>{(value) => <div>{value}</div>}</Ctx.Consumer>;

const NotWorking = () => (
    <Ctx.Provider value="works">
        <ClassConsumer />
    </Ctx.Provider>
);

const Working = () => (
    <Ctx.Provider value="works">
        <FuncConsumer />
    </Ctx.Provider>
);

console.log(render(<NotWorking />));
console.log(render(<Working />));
```

Dont pay any attention to `React` imports, its aliased to `preact` in webpack conf.

w4zZz4p:
Closes #121

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -92,14 +92,18 @@ function renderToString(vnode, context, opts, inner, isSvgMode, selectValue) {
 			}
 			else {
 				// class-based components
+				let cxType = nodeName.contextType;
+				let provider = cxType && context[cxType.__c];
+				let cctx = cxType != null ? (provider ? provider.props.value : cxType._defaultValue) : context;
+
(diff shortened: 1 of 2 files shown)


marvinhagemeister:
Released as `5.1.0` :tada:

PR #124, opened 2019-11-01
Fix effects breaking in SSR due to missing _renderCallbacks

developit:
/cc @marvinhagemeister


marvinhagemeister:
I upgraded to Preact 10.0.4 because the issue is not present on older versions. Moved the fix a bit around so that it is applied for functional components.

PR #130, opened 2019-12-18
Fix support for default context value

loklaan:
Issue here was that the [mangled `_defaultValue`](https://github.com/preactjs/preact/blob/71425ff63ddd59989a7c657f659b70960d457600/mangle.json#L36) prop wasn't being used. 😱

Pretty edge case for _most_ libraries I suppose, which is why it might not have been caught yet

### Checklist

- [x] Added test


marvinhagemeister:
Published as `5.1.3` :tada:

loklaan:
Yaaas - quick publish! ty

PR #131, opened 2020-01-22
Add TypeScript definitions for JSX renderer

squidfunk:
This PR adds the TypeScript definitions for `preact-render-to-string/jsx`.

### DIFF
--- a/src/jsx.d.ts
+++ b/src/jsx.d.ts
@@ -1,0 +1,13 @@
+import { VNode } from 'preact';
+
+interface Options {
+  jsx?: boolean;
+  xml?: boolean;
+  functions?: boolean
+  functionNames?: boolean,
+  skipFalseAttributes?: boolean
+  pretty?: boolean | string;
+}
+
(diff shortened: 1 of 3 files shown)


squidfunk:
For some strange reason Pika CI fails. Let me know if I can help.

src/jsx.d.ts
```diff
@@ -0,0 +1,13 @@
+import { VNode } from 'preact';
```
marvinhagemeister: For any definition files other than the one referenced in `package.json` TypeScript expects them to be next to the real code. In our case `jsx.js` is referenced from a file of the same name in the root of this repository. TypeScript is unable to understand that indirection. Moving `jsx.d.ts` to the repo root will fix this 👍

marvinhagemeister:
We're working on the pika error. Not sure why it fails but we only recently integrated it and we don't really use it as a CI despite the name. If Travis is green than it's all good 💯

src/jsx.d.ts
squidfunk: You're right. Give me a sec...

PR #133, opened 2020-01-29
Export renderToString as renderToStaticMarkup

sventschui:
Export renderToString as renderToStaticMarkup

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -256,6 +256,7 @@ export default renderToString;
 
 export {
 	renderToString as render,
+	renderToString as renderToStaticMarkup,
 	renderToString,
 	shallowRender
 };


src/index.js
```diff
@@ -256,6 +256,7 @@ export default renderToString;
 
 export {
 	renderToString as render,
+	renderToString as renderToStaticMarkup,
```
marvinhagemeister: Super minor nit: Maybe we can add a little test here that checks if it's exported?

https://github.com/preactjs/preact-render-to-string/blob/master/test/index.js

sventschui: Haha I hoped nobody requests this 😂 I‘ll add one 🙈

marvinhagemeister: If the branch wouldn't be in a fork I'd add them myself :see_no_evil:

PR #134, opened 2020-01-30
Add submodule exports

### LINKED ISSUE #144: preact hooks SSR doesn't work in MJS
### Reproduction

https://github.com/dfabulich/preact-hooks-ssr-esm

It's a simple repo depending on `preact`, `preact-render-to-string`, and `htm`, with two scripts, `index.js` (using CommonJS) and `index.mjs` (using ES modules), that are designed to be two versions of the same server-side rendering hooks sample.

I'm using Node v14.1.0.

### Steps to reproduce

Check out that repo on github and `npm install`. You can run `node index` and it will print `<div class="foo">0</div>`. Now run `node index.mjs`.

### Expected Behavior

It should print `<div class="foo">0</div>` just like the CJS version does

### Actual Behavior

```
Error: Hook can only be invoked from render methods.
    at Object.n.__h (file:///tmp/preact-hooks-ssr-esm/node_modules/preact/debug/dist/debug.mjs:1:5847)
    at v (file:///tmp/preact-hooks-ssr-esm/node_modules/preact/hooks/dist/hooks.mjs:1:117)
    at p (file:///tmp/preact-hooks-ssr-esm/node_modules/preact/hooks/dist/hooks.mjs:1:272)
    at m (file:///tmp/preact-hooks-ssr-esm/node_modules/preact/hooks/dist/hooks.mjs:1:241)
    at Object.App (file:///tmp/preact-hooks-ssr-esm/index.mjs:10:28)
    at g (/tmp/preact-hooks-ssr-esm/node_modules/preact-render-to-string/dist/index.js:1:2233)
    at file:///tmp/preact-hooks-ssr-esm/index.mjs:14:13
    at ModuleJob.run (internal/modules/esm/module_job.js:110:37)
    at async Loader.import (internal/modules/esm/loader.js:179:24)
```

marvinhagemeister:
In Node v13, package exports were introduced as an official way to define sub-paths/modules within your package: https://nodejs.org/api/esm.html#esm_package_exports

Fixes #144 .

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,8 +7,21 @@
   "umd:main": "dist/index.js",
   "module": "dist/index.module.js",
   "jsnext:main": "dist/index.module.js",
+  "exports": {
+    ".": {
+      "require": "./dist/index.js",
+      "import": "./dist/index.mjs",
+      "browser": "./dist/index.module.js"
+    },
+    "./jsx": {
+      "require": "./dist/jsx.js",
+      "import": "./dist/jsx.mjs",
+      "browser": "./dist/jsx.module.js"
+    }
+  },
   "scripts": {
     "build": "npm run -s transpile && npm run -s transpile:jsx && npm run -s copy-typescript-definition",
+    "postbuild": "node ./config/node-13-exports.js",
     "transpile": "microbundle src/index.js -f es,umd --target web --external preact",
     "transpile:jsx": "microbundle src/jsx.js -o dist/jsx.js --target web --external none && microbundle dist/jsx.js -o dist/jsx.js -f cjs",
     "copy-typescript-definition": "copyfiles -f src/*.d.ts dist",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -7,6 +7,13 @@
   "umd:main": "dist/index.js",
   "module": "dist/index.module.js",
   "jsnext:main": "dist/index.module.js",
+  "exports": {
```
JoviDeCroock: Let's add `.` as well like we do in [preact](https://github.com/preactjs/preact/blob/master/package.json#L13)

marvinhagemeister: Good call, updated that 👍

marvinhagemeister:
@JoviDeCroock would you mind reviewing it again? Turns out that this PR fixes #144 .

JoviDeCroock:
@marvinhagemeister mergeable after adding these two https://github.com/preactjs/preact/blob/master/package.json#L52-L53 this fails in rollup otherwise since they look at `package.json`

marvinhagemeister:
@JoviDeCroock Good catch, completely missed those! There are a lot of things to be aware of with node's new `export` declaration it seems like 🙈

PR #140, opened 2020-04-01
Remove unnecessary line

38elements:
This may be breaking change.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -303,6 +303,5 @@ export default renderToString;
 
 export {
 	renderToString as render,
-	renderToString,
 	shallowRender
 };
--- a/src/index.js
+++ b/src/index.js
@@ -303,5 +303,6 @@ export default renderToString;
 
 export {
 	renderToString as render,
+	renderToString,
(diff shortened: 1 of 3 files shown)


src/index.js
```diff
@@ -303,6 +303,5 @@ export default renderToString;
 
 export {
 	renderToString as render,
-	renderToString,
```
38elements, line 306: This is not in document, test, type definition. 

https://github.com/preactjs/preact-render-to-string/blob/552756f827f4628489e7356a4c0b56f14d074b31/src/index.d.ts#L1-L11

https://github.com/preactjs/preact-render-to-string/blob/552756f827f4628489e7356a4c0b56f14d074b31/test/index.js#L4-L19

marvinhagemeister:
Yeah that's a breaking change like you noted and I'm not sure if it's warranted. Maybe we can avoid that by just updating our type definition instead? What do you think?

38elements:
I think it is resolved by adding renderToString to type definition.

PR #150, opened 2020-05-08
Introduce github actions workflows

sventschui:
This introduces two workflows. Open for feedback and improvements.

**push.yml**

A regular CI build that runs tests.

Todo
- [ ] Test coverage

**release-publish.yml**

Triggers when a release is published on github. It runs `npm version <tag-name>` and attaches the npm package tarball to the release. It then pushes the commit created by npm version and force pushes the release tag to the same commit.

- It currently can not run `npm publish` preact packages require 2FA for publishing
- It will fail and not update the branch/commit when there were commits added to the referenced branch in the meantime (kinda "expected behaviour")
- It will fail if the release was not created on a branch but rather a commit (SHA)

### DIFF
--- a/.github/workflows/push.yml
+++ b/.github/workflows/push.yml
@@ -1,0 +1,33 @@
+name: CI
+
+on:
+    push:
+        branches:
+            - "*"
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Use Node.js 12
+      uses: actions/setup-node@v1
+      with:
+        node-version: 12.x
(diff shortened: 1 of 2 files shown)


developit:
@sventschui I've added `{{ secrets.PREACT_NPM_TOKEN }}`.

sventschui:
Awesome 🥳 Only thing left then is how to handle the protected branches as pointed out by @marvinhagemeister. I guess it would be best to adjust the workflow to create a PR and then do the release on PR merge

rschristian:
Going to close this on out as both uses were eventually supported via #179 & #204.

PR #154, opened 2020-06-02
Add option to preserve whitespace for certain elements

### LINKED ISSUE #143: Pretty rendering breaks <pre> tags
When setting `{pretty: true}` in options, looks like any `<pre>` tags in the output get extra indentation added (based on the nesting level of the tag), breaking its contents.

sventschui:
Add an option to preserve whitespace/indentation for certain elements. Defaults to `['pre']`

**Caution** I'm not 100% sure this doesn't break pretty printing, tests seem fine though.

Fixes #143


src/index.js
```diff
@@ -73,7 +74,7 @@ function renderToString(vnode, context, opts, inner, isSvgMode, selectValue) {
 			for (let i = 0; i < children.length; i++) {
 				rendered += (i > 0 && pretty ? '\n' : '') + renderToString(children[i], context, opts, opts.shallowHighOrder!==false, isSvgMode, selectValue);
 			}
-			return rendered;
+			return pretty ? indent(rendered, indentChar) : rendered;
```
sventschui, line 77: Had to move the call to indent here to not indent HTML content but do indent Fragment children. I'm not 100% sure this doesn't break other things. I'll try to add some more tests

src/index.js
```diff
@@ -20,6 +20,7 @@ const noop = () => {};
  *	@param {Boolean} [options.shallow=false]	If `true`, renders nested Components as HTML elements (`<Foo a="b" />`).
  *	@param {Boolean} [options.xml=false]		If `true`, uses self-closing tags for elements without children.
  *	@param {Boolean} [options.pretty=false]		If `true`, adds whitespace for readability
+ *	@param {Boolean} [options.preserveWhitespace=['pre']]		Array of HTML tags for which to preserve indentation.
```
marvinhagemeister, line 23: Nit: type should be string array here 👍

sventschui:
It seems the change will break indentation of text nodes. Honestly I'm not sure we can actually change text indent in a safe way. I'd therefore suggest to not indent html (dangerouslySetInnerHtml, string children, etc.) in pretty mode

sventschui:
Just to keep you posted: I'll add the discussion with @marvinhagemeister on slack here publicly:

Marvin mentioned the issue is that we are indenting after rendering instead of during rendering. While trying to change that I ran into issues on how to determine whether we need to wrap/indent text content/children.

natemoo-re:
This is unfortunately a pretty big issue for [`microsite`](https://github.com/natemoo-re/microsite), a Preact-based SSG, since developer blogs with code examples are a big use case. I'd like to keep `pretty` enabled in the spirit of the "view source" web.

Any updates?

developit:
@natemoo-re Are you using the `{ pretty: true }` option? It's really only designed for use in things like tests - none of the options for renderToString() are part of the public API.

natemoo-re:
@developit I was at the time! We've since moved to formatting the output as a separate step for other reasons.

developit:
@natemoo-re ah okay, good. I'd like to remove this from the default render to string implementation.

PR #156, opened 2020-06-02
Add voidElements config option

### LINKED ISSUE #52: allow custom void elements
We have the case that we want to render more elements which can be also void (XML syntax).

It would be good to allow this with an option.
Reference: https://github.com/developit/preact/issues/919

sventschui:
Fixes #52

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -6,7 +6,7 @@ const SHALLOW = { shallow: true };
 // components without names, kept as a hash for later comparison to return consistent UnnamedComponentXX names.
 const UNNAMED = [];
 
-const DEFAULT_VOID_ELEMENTS = /^(area|base|br|col|embed|hr|img|input|link|meta|param|source|track|wbr)$/;
+const VOID_ELEMENTS = /^(area|base|br|col|embed|hr|img|input|link|meta|param|source|track|wbr)$/;
 
 const noop = () => {};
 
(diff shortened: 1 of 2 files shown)


src/index.js
```diff
@@ -219,7 +220,7 @@ function renderToString(vnode, context, opts, inner, isSvgMode, selectValue) {
 	s = `<${nodeName}${s}>`;
 	if (String(nodeName).match(/[\s\n\\/='"\0<>]/)) throw new Error(`${nodeName} is not a valid HTML tag name in ${s}`);
 
-	let isVoid = String(nodeName).match(VOID_ELEMENTS);
+	let isVoid = String(nodeName).match(opts.voidElements || DEFAULT_VOID_ELEMENTS);
```
marvinhagemeister, line 223: This PR is awesome :raised_hands: 

We can combine the two by always checking for `DEFAULT_VOID_ELEMENTS` and checking for `opts.voidElements` if it was defined. Otherwise defining a custom element requires the user to add all default void elements in their regex.

sventschui: I first had 2 config options (`voidElements` and `additionalVoidElements`) where the later did exactly what you mentioned. With the former one has the full flexibility with the drawback you mentioned. Not sure what is best here.

JoviDeCroock: We could export the default_void_elements, this way the user has access to them and can expand on them.

marvinhagemeister: Or something like:

```js
let isVoid = nodeName.match(opts.voidElements) || nodeName.match(DEFAULT_VOID_ELEMENTS)
```

sventschui: @marvinhagemeister the only drawback I saw there is that one is not able to completely opt out of void elements. But that might be a too crazy usecase. 

I‘ll adjust the PR soonish

marvinhagemeister: Seeing that Preact's main focus is the web, I think we're good with that :+1:

PR #164, opened 2020-08-02
feat(index): add condition encode

zhijs:
hi～，I was use this code in my web page, when i run code like this:
```javascript
const preactRenderToString = require('preact-render-to-string')
const preact = require('preact')

var vnode = preact.createElement("style", {
  type: "text/style"
}, `
  .container::before {
    content: "";
    width: 0;
    height: 0;
    overflow: hidden;
  }`);
console.log(preactRenderToString(vnode));
console.log(vnode)
```
and i got output that
```html
<style type="text/style">
  .container::before {
    content: &quot;&quot;;
    width: 0;
    height: 0;
    overflow: hidden;
  }</style>
```
and the style will not work, i think it is for xss defense， but  I presume the style don't have to encode, that is why I pull this request.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -10,6 +10,7 @@ const VOID_ELEMENTS = /^(area|base|br|col|embed|hr|img|input|link|meta|param|sou
 
 const noop = () => {};
 
+const skillEncodeNodeTypes = ['style']
 
 /** Render Preact JSX + Components to an HTML string.
  *	@name render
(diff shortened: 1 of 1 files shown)


sventschui:
Thanks for your PR. I'm on my phone so sorry for the short answer. 

This is expected behaviour in JSX land, as you mentioned to prevent accidental XSS. 

I think you're looking for dangerouslySetInnerHTML, see https://github.com/preactjs/preact-render-to-string/blob/694155845d7d98637d80f2fd10b92e64412cfe50/test/render.js#L627 for am example.

Dows this solve your usecase?

zhijs:
> Thanks for your PR. I'm on my phone so sorry for the short answer.
> 
> This is expected behaviour in JSX land, as you mentioned to prevent accidental XSS.
> 
> I think you're looking for dangerouslySetInnerHTML, see
> 
> https://github.com/preactjs/preact-render-to-string/blob/694155845d7d98637d80f2fd10b92e64412cfe50/test/render.js#L627
> 
> for am example.
> Dows this solve your usecase?

thank you very much, it solve my problem

Issue #53, opened 2017-11-04
Allow override of `VOID_ELEMENTS`


malchata:
In some instances, it's not desirable for certain tags to be treated exclusively as self-closing elements, such as with the `<link>` tag in RSS feeds. It might be useful to allow an additional option in the `renderToString` method to allow users to supply an array of elements they want (or as in this case, don't want) to be treated as self-closing. This may not really be used very much, but in cases where XML is generated, it could be a nice little feature to add.

If you're down, I'm happy to add it. I also see that this has been referenced in a somewhat different way in #52.

developit:
Yup. Sortof a duplicate, but it should be fine.

TBH I know it's ugly, but we could export the global list..

developit:
Let's consolidate on #52.

PR #166, opened 2020-10-08
Override default void elements

gpoitch:
When working with xml, it is undesirable to have the default html void elements closed (e.g. `link`). See #53
The `voidElements` option added in #156 is additive to the default elements.

This PR makes defining a `voidElements` option become the exclusive void elements.  It also exposes the default html void elements as a named export if you want to add on to it.

Technically a breaking change, so open to feedback on how to handle.


gpoitch:
Closing in favor of https://github.com/preactjs/preact-render-to-string/pull/167

PR #167, opened 2020-10-08
[bugfix] do not close void elements with a closing tag

gpoitch:
If an element that is typically a void element is being used as a standard element with children, don't self close it.
The existing test for this was not doing what its description said.  This is required for better xml support. See #53


marvinhagemeister:
@gpoitch I'm not familiar enough with HTML to know this: Can XML-tags be of type void and have children at the same time?

gpoitch:
@marvinhagemeister XML could be anything as long as it's properly closed.

Currently, something like this using a tag in the void list will become:
```<link>foo.com</link>``` becomes `<link />foo.com`

which I don't think is desirable regardless of xml or html context.  

This simply lets the existence of children opt-out of creating the void element.

marvinhagemeister (approved):
Awesome, thank you so much for your PR 🙌

PR #170, opened 2020-10-15
Optimize HTML entity encoding

developit:
See benchmarks:
https://esbench.com/bench/5f88af6cb4632100a7dcd414

`Scan (str) [3]` is this revised implementation, which is 50% faster than the previous implementation (`replace`).


JoviDeCroock (approved):
Nice one!

marvinhagemeister (approved):

PR #171, opened 2020-10-17
utilize stringbuilder in recursion

JoviDeCroock:
This avoids relying too much on "trampolining", rather than relying on return values we'll mutate an outer array.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -81,15 +81,14 @@ function renderToString(vnode, context, opts, inner, isSvgMode, selectValue, out
 		isComponent = true;
 		if (opts.shallow && (inner || opts.renderRootComponent === false)) {
 			nodeName = getComponentName(nodeName);
-		} else if (nodeName === Fragment) {
-			let rendered = '';
+		}
+		else if (nodeName===Fragment) {
 			let children = [];
 			getChildren(children, vnode.props.children);
 
 			for (let i = 0; i < children.length; i++) {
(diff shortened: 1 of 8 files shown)


src/index.js
```diff
@@ -305,19 +339,22 @@ function renderToString(vnode, context, opts, inner, isSvgMode, selectValue) {
...
-						selectValue
-					);
+					nodeName === 'svg'
+						? true
+						: nodeName === 'foreignObject'
+						? false
+						: isSvgMode;
+
+				// TODO: we can't rely on this return-value...
+				let ret = renderToString(
```
JoviDeCroock, line 349: Not quite sure how to handle this, we could try utilizing a custom array as last argument that scopes the result to this child but then again we merge adjacent nodes from lastpass, ...

marvinhagemeister: The only solution I can come up with is to track the iteration via a simple stack and push some sort of marker to that for when a node should be closed. Those can be stored in a second arrays and popped whenever a marker is encountered in the "work" stack.

```js
let output = []
let stack = [vnode]
let closing = [];
let item;
while ((item = stack.pop()) !== undefined) {
  // If we encounter a marker we need to close the element
  if (item === null) {
    output.push(closing.pop())
    continue;
  }

  output.push(`<${item.name}>`, null); // <-- push marker
  closing.push(`</${item.name}>`);

  stack.push(...item.children);
}
```

developit:
@JoviDeCroock I just realized this is one of the variants I ended up testing in my other benchmarks PR - I think if there's a way we can get the render times to be comparable with the argument-passed Array stringbuilder approach I'd like that since it avoids relying on globals (which breaks future async stuff). I just couldn't get the performance to match the global Ropes-based string implementation.

PR #172, opened 2020-10-21
Invoke _diff hook

sventschui:
As otherwise `preact/debug` throws errors when using hooks as `hooksAllowed` is never set to `true`.

I could not yet come up with a test repro for this though!


marvinhagemeister (approved):
Good catch :+1:

PR #173, opened 2020-10-24
Prefer ESM over CJS in package exports

developit:


marvinhagemeister (approved):

balazsorban44:
Sorry for posting this here, but I suspect that this change broke our implementation and people started reporting bugs because of this. See https://github.com/nextauthjs/next-auth/issues/1470

Is this anything wrong with our import here: https://github.com/nextauthjs/next-auth/blob/main/src/server/pages/signin.js#L2

marvinhagemeister:
@balazsorban44 Just published `v5.1.14` which reverts back to the previous commonjs export behaviour. Looks like something changed in a transient dependency. Likely the previous versions were done with an older `node_modules` version that didn't match the newest packages specified in `package.json`.

PR #174, opened 2020-10-31
Prevent to self-close if dangerouslySetInnerHTML set

clomie:
This change prevents to self-close if the element has `dangerouslySetInnerHTML` prop. This is required for better xml support.

Currently behavior:
```
import { h } from 'preact'
import renderToString from 'preact-render-to-string'

console.log(renderToString(<a dangerouslySetInnerHTML={{ __html: '<b />' }}></a>, undefined, { xml: true }))
// <a><b / />

console.log(renderToString(<link dangerouslySetInnerHTML={{ __html: '<foo />' }}></link>))
// <link><foo / />
```

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -351,13 +351,13 @@ function renderToString(vnode, context, opts, inner, isSvgMode, selectValue) {
 		}
 	}
 
-	if (pieces.length) {
+	if (pieces.length || html) {
 		s += pieces.join('');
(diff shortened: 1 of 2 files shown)


sventschui (approved):
Thank you very much for your contribution! This fix looks good to me, just waiting for another approver before merging.

JoviDeCroock (approved):
Looks great, thank you!

PR #185, opened 2021-03-10
Remove all default exports (major)

marvinhagemeister:
Mixing both default and named exports has no clearly defined behaviour and can easily lead to unexpected results. Some tools use a combination of "default" + "__esModule" property and others ignore named exports completely if both are found. Those issues are difficult to debug when multiple tools are used together.
    
This change makes some code redundant by not having to patch a hand crafted exports object for commonjs environments.


JoviDeCroock (approved):

developit:
Couldn't we provide only a default export? None of the other exported methods were widely used.

PR #190, opened 2021-03-15
Improve string rendering

developit:
Depends on #189.

**Update: I'm silly, this is basically neutral and a bug was just disabling encoding.**


marvinhagemeister:
> Update: I'm silly, this is basically neutral and a bug was just disabling encoding.

@developit does that mean that this PR can be closed?

developit:
@marvinhagemeister I'm going to re-test.

PR #192, opened 2021-03-20
Performance improvements!

developit:
Screen Shot 2021-03-19 at 8 25 11 PM

Some benchmarks, for reference:
https://esbench.com/bench/5f88af6cb4632100a7dcd414

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -6,7 +6,7 @@ import {
 	assign,
 	getChildren
 } from './util';
-import { options } from 'preact';
+import { options, Fragment } from 'preact';
 
 /** @typedef {import('preact').VNode} VNode */
 
@@ -68,7 +68,6 @@ function _renderToString(vnode, context, opts, inner, isSvgMode, selectValue) {
 	let pretty = opts.pretty,
 		indentChar = pretty && typeof pretty === 'string' ? pretty : '\t';
 
-	// wrap array nodes in Fragment
 	if (Array.isArray(vnode)) {
 		let rendered = '';
 		for (let i = 0; i < vnode.length; i++) {
@@ -94,6 +93,17 @@ function _renderToString(vnode, context, opts, inner, isSvgMode, selectValue) {
 		isComponent = true;
 		if (opts.shallow && (inner || opts.renderRootComponent === false)) {
 			nodeName = getComponentName(nodeName);
+		} else if (nodeName === Fragment) {
+			const children = [];
+			getChildren(children, vnode.props.children);
+			return _renderToString(
+				children,
+				context,
+				opts,
+				opts.shallowHighOrder !== false,
+				isSvgMode,
+				selectValue
+			);
 		} else {
 			let rendered;
 
--- a/src/index.js
+++ b/src/index.js
@@ -6,7 +6,9 @@ import {
 	assign,
 	getChildren
 } from './util';
-import { options, Fragment, createElement } from 'preact';
+import { options } from 'preact';
+
+/** @typedef {import('preact').VNode} VNode */
 
 const SHALLOW = { shallow: true };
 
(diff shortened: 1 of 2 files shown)


src/index.js
```diff
@@ -56,48 +60,40 @@ function _renderToString(vnode, context, opts, inner, isSvgMode, selectValue) {
...
-					_renderToString(
-						children[i],
-						context,
-						opts,
-						opts.shallowHighOrder !== false,
-						isSvgMode,
-						selectValue
-					);
-			}
-			return rendered;
```
marvinhagemeister, line 100: I think we should keep this branch here as a fast path for Fragments. It allows us to avoid allocating a component instance that we don't need since Fragments can't be stateful. This branch is hit for cases like this one:

```jsx
<div>
  <Fragment>
    <span />
  <Fragment />
</div>
```

src/util.js
```diff
@@ -1,17 +1,12 @@
...
-	'&': '&amp;',
-	'<': '&lt;',
-	'>': '&gt;',
-	'"': '&quot;'
-};
-const replaceTag = (tag) => tagsToReplace[tag] || tag;
 export function encodeEntities(s) {
-	if (typeof s !== 'string') s = String(s);
-	return s.replace(HTML_ENTITY_REG, replaceTag);
+	return String(s)
```
marvinhagemeister, line 5: Let's add a comment to summarize our findings from yesterdays quick benchmarking session.

src/index.js
```diff
@@ -56,48 +60,40 @@ function _renderToString(vnode, context, opts, inner, isSvgMode, selectValue) {
 		return '';
 	}
 
+	// #text nodes
+	if (typeof vnode !== 'object') {
```
marvinhagemeister, line 64: This is great! Makes sense to move it up 👍

src/index.js
```diff
@@ -56,48 +60,40 @@ function _renderToString(vnode, context, opts, inner, isSvgMode, selectValue) {
...
+		return encodeEntities(vnode);
+	}
+
+	let pretty = opts.pretty,
+		indentChar = pretty && typeof pretty === 'string' ? pretty : '\t';
+
 	// wrap array nodes in Fragment
 	if (Array.isArray(vnode)) {
-		vnode = createElement(Fragment, null, vnode);
+		let rendered = '';
```
marvinhagemeister, line 73: +1 For inlining Fragments here 👍

marvinhagemeister (commented):
This is great! Left a few minor comments. Let me know what you think 👍

src/util.js
JimLiu: I've always felt that this change was suspicious and could be the root cause of the performance issues I reported previously for versions after 5.1.2.

I plan to deploy a test release next Monday and roll back this part of the change separately from the 5.1.16 version of the code to see if there are still performance issues.

src/index.js
```diff
@@ -215,7 +211,7 @@ function _renderToString(vnode, context, opts, inner, isSvgMode, selectValue) {
 				continue;
 			}
 
-			if (name.match(/[\s\n\\/='"\0<>]/)) continue;
+			if (UNSAFE_NAME.test(name)) continue;
```
JimLiu, line 214: 👍

src/util.js
JimLiu: FYI
image

src/index.js
developit: good point - added!

src/util.js
developit: @JimLiu - that benchmark website produces statistically meaningless results. It's effectively doing `while(performance.now()-start < 5000) test()`, which makes the order of benchmarks matter and doesn't reflect real JS VM workloads.

The only current benchmarking website that produces statistically meaningful results is [ESbench](https://esbench.com) (disclaimer: I maintain it). The reason for this is because it uses the same [Benchmark.js](https://benchmarkjs.com/) (the library that powered JSPerf, built by Matthias Bynens from the V8 team). Benchmark.js performs JIT warmup runs, encapsulates benchmarks in functions that prevent synthetic VM optimizations (like hoisting/inlining of benchmark code), and returns median values with standard deviation.

Here's the benchmarks we use for `encodeEntities` on ESBench:
https://esbench.com/bench/5f88af6cb4632100a7dcd414

Screen Shot 2021-03-25 at 11 45 36 AM

marvinhagemeister (approved):

Issue #191, opened 2021-03-18
Performance issue with 5.1.15


JimLiu:
I reported a performance issue last time #180, the issue had been resolved after we rolled back to `v5.1.2`

this week we deployed the new version `v5.1.15`, got performance issue again.
![image](https://user-images.githubusercontent.com/648674/111706776-dfc4c100-8810-11eb-9473-760e1a59ace9.png)

and got back to normal after rolled back to `v5.1.2`

I checked the code between `v5.1.2` to `v5.1.15` [415a377..9a5e748](https://github.com/preactjs/preact-render-to-string/compare/415a377..9a5e748#diff-9c0ec1b0a889e599f3ff81590864dd9dc65684a86b41f1da75acc53fafed12e3R7), memory leak?!

The code is running on the server side!

![image](https://user-images.githubusercontent.com/648674/111709436-99259580-8815-11eb-9b04-960e6ced7451.png)

developit:
@JimLiu This performance issue was addressed in `5.1.16`, which does not do any memoization. Are you able to try that version?

https://github.com/preactjs/preact-render-to-string/releases/tag/v5.1.16

JimLiu:
> @JimLiu This performance issue was addressed in `5.1.16`, which does not do any memoization. Are you able to try that version?
> 
> https://github.com/preactjs/preact-render-to-string/releases/tag/v5.1.16

Thank you for your reply, will try 5.1.16 tomorrow

developit:
Awesome! We're working on landing some performance improvements - perhaps we could ping you when they're ready to test and get your feedback prior to release?

JimLiu:
Sure, preact render is currently a bottleneck for our performance.

We'd be happy to see a significant performance improvement with it as well. Please let me know when the new version is ready and I'll give you feedback.

JimLiu:
@developit 
We deployed v5.1.16 to production, it's better than v5.1.15 for sure, but you can still see a significant increase in CPU at some period of time. I'll need more time to keep observing.

![image](https://user-images.githubusercontent.com/648674/111828816-da6f8100-88b9-11eb-84ad-1e703a2909fc.png)

JimLiu:
The performance of v5.1.16 is not as good as that of v5.1.2
![image](https://user-images.githubusercontent.com/648674/111841977-ced98580-88cc-11eb-8081-b87bf8f4d893.png)

JimLiu:
[...]

PR #193, opened 2021-03-24
Skip effects during RTS

JoviDeCroock:
This skips effects during `renderToString` just like React, as this function runs synchronously anyway and isn't subject to effectful rerenders.

Fixes: https://github.com/preactjs/preact-render-to-string/issues/191

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -14,7 +14,6 @@ const SHALLOW = { shallow: true };
 
 // components without names, kept as a hash for later comparison to return consistent UnnamedComponentXX names.
 const UNNAMED = [];
-const _skipEffects = '__s';
 
 const VOID_ELEMENTS = /^(area|base|br|col|embed|hr|img|input|link|meta|param|source|track|wbr)$/;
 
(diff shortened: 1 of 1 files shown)


src/index.js
```diff
@@ -12,6 +12,7 @@ const SHALLOW = { shallow: true };
 
 // components without names, kept as a hash for later comparison to return consistent UnnamedComponentXX names.
 const UNNAMED = [];
+const _skipEffects = "__s";
```
marvinhagemeister, line 15:
```suggestion
const _skipEffects = '__s';
```

src/index.js
```diff
@@ -43,10 +44,13 @@ let shallowRender = (vnode, context) => renderToString(vnode, context, SHALLOW);
 
 const EMPTY_ARR = [];
 function renderToString(vnode, context, opts) {
+	const previousSkipEffects = options[_skipEffects];
+	options[_skipEffects] = true;
```
developit, line 48: Do we know if Terser inlines this? I'm 99% sure it will, just we're using the minified properties everywhere else.

developit (approved):

src/index.js
marvinhagemeister: Pushed a commit that inlines it.

PR #199, opened 2021-07-30
Transform attribute names to proper casing

### LINKED ISSUE #18: srcSet not converted to srcset
This works correctly in preact proper

gpoitch:
Makes sure attributes are converted to their proper casing when rendering to string. There was a similar attempt with #55 but it was never completed.  Reviving it particularly because Safari is excessively downloading an img's src upon hydration due to `srcSet` casing mismatch between preact/preact-render-to-string. This also adds better support for attributes specific to SVG/RSS/XML. 

Incorporated the concerns from the comments in the other attempts/issues.

Attribute list reference: https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes
React reference: https://github.com/facebook/react/blob/main/packages/react-dom/src/shared/possibleStandardNames.js

Closes #55 
Closes #18


marvinhagemeister:
@gpoitch I think it's definitely something we should address. Regarding the closing: Did you ran into problems with this PR? I'm mainly curious to see if there is a way without changing the casing in the JSX renderer, but lowercase it in the default mode.

gpoitch:
@marvinhagemeister accidental auto-close referencing this PR in a private repo.  Not familiar with default vs jsx mode - I'll have to read the source code more and circle back.

gpoitch:
@marvinhagemeister I updated it so the jsx renderer is unaffected by applying the normalization after the attribute hook intercepts the standard rendering.

src/index.js
```diff
@@ -17,6 +17,12 @@ const UNNAMED = [];
 
 const VOID_ELEMENTS = /^(area|base|br|col|embed|hr|img|input|link|meta|param|source|track|wbr)$/;
 
+const DASHED_ATTRS = /^(acceptC|httpE|(clip|color|fill|font|glyph|marker|stop|stroke|text|vert)[A-Z])/;
+const CAMEL_ATTRS = /^(isP|viewB)/;
```
developit, line 21: What's `isP` matching here?

src/index.js
```diff
@@ -426,6 +435,19 @@ function getFallbackComponentName(component) {
...
+	if (CAMEL_ATTRS.test(name)) return name;
+
+	if (DASHED_ATTRS.test(name))
+		return name.replace(CAPITAL_REGEXP, (w) => '-' + w.toLowerCase());
+
+	if (COLON_ATTRS.test(name))
+		return name.replace(CAPITAL_REGEXP, (w) => ':' + w.toLowerCase());
+
+	return name.toLowerCase();
+}
```
developit, line 449: HTML allows mixed-case attribute names, so we shouldn't default to converting things to lowercase.

```suggestion
function getAttributeNameInHtmlCase(name) {
	if (DASHED_ATTRS.test(name)) {
		return name.replace(CAPITAL_REGEXP, '-$1');
	}
	return name.replace(COLON_ATTRS, '$1:$2');
}
```

src/index.js
```diff
@@ -17,6 +17,12 @@ const UNNAMED = [];
 
 const VOID_ELEMENTS = /^(area|base|br|col|embed|hr|img|input|link|meta|param|source|track|wbr)$/;
 
+const DASHED_ATTRS = /^(acceptC|httpE|(clip|color|fill|font|glyph|marker|stop|stroke|text|vert)[A-Z])/;
+const CAMEL_ATTRS = /^(isP|viewB)/;
+const COLON_ATTRS = /^(xlink|xml|xmlns)[A-Z]/;
```
developit, line 22: (see below)

```suggestion
const COLON_ATTRS = /^(xlink|xml|xmlns)([A-Z])/;
```

src/index.js
```diff
@@ -281,6 +287,8 @@ function _renderToString(vnode, context, opts, inner, isSvgMode, selectValue) {
 				// <textarea value="a&b"> --> <textarea>a&amp;b</textarea>
 				propChildren = v;
 			} else if ((v || v === 0 || v === '') && typeof v !== 'function') {
+				name = getAttributeNameInHtmlCase(name);
```
developit, line 290: Likely faster if this is done on L310:

```diff
-				s += ` ${name}="${encodeEntities(v)}"`;
+				s += ` ${getAttributeNameInHtmlCase(name)}="${encodeEntities(v)}"`;
```

src/index.js
gpoitch: isPermaLink (RSS)

src/index.js
gpoitch: Need the transformed name for the `if (v === true || v === '')` block

src/index.js
gpoitch: Updated the regex shorthands.
Updated the function name - it's not specific to HTML. Using it for HTML/RSS/XML/SVG.
Without lowercasing by default I think there is going to have to be a big whitelist regex.

JoviDeCroock (approved):

src/index.js
```diff
@@ -17,6 +17,12 @@ const UNNAMED = [];
 
 const VOID_ELEMENTS = /^(area|base|br|col|embed|hr|img|input|link|meta|param|source|track|wbr)$/;
 
+const DASHED_ATTRS = /^(acceptC|httpE|(clip|color|fill|font|glyph|marker|stop|stroke|text|vert)[A-Z])/;
+const CAMEL_ATTRS = /^(isP|viewB)/;
+const COLON_ATTRS = /^(xlink|xml|xmlns)([A-Z])/;
+
+const CAPITAL_REGEXP = /([A-Z])/g;
+
 const UNSAFE_NAME = /[\s\n\\/='"\0<>]/;
```
SukkaW, line 26: Can they cover all the attributes? `react-dom` has an entire file for `possibleStandardNames`:

https://github.com/facebook/react/blob/ca106a02d1648f4f0048b07c6b88f69aac175d3c/packages/react-dom/src/shared/possibleStandardNames.js

If they can, at least add a unit test case to cover them all.

PR #207, opened 2021-12-12
Add test case for attribute name transformation (#199)

SukkaW:
This PR adds a full test case set for PR #199.

@gpoitch Would you mind including the change in your PR?


SukkaW:
> Could be a good idea. That list from React isn't quite usable as-is. I think they are using it for other purposes but we can derive a proper list from that.
> 
> It has a bunch of duplicates values e.g.:
> 
> ```
> acceptcharset: 'acceptCharset',
> 'accept-charset': 'acceptCharset'
> ```
> 
> And others that aren't attributes e.g. `innerHTML`.

Oh, you are right.

So I look into `react-dom` again: https://github.com/facebook/react/blob/c88fb49d37fd01024e0a254a37b7810d107bdd1d/packages/react-dom/src/__tests__/ReactDOMAttribute-test.js

But the test cases seem to be limited and might not cover all possible case.

gpoitch:
Here are some attribute lists that appear to be maintained:
https://github.com/wooorm/html-element-attributes
https://github.com/wooorm/svg-element-attributes

PR #209, opened 2021-12-15
fix: On empty className w/ compat, class attr will no longer be duplicated

### LINKED ISSUE #208: Empty class attribute duplicated with preact/compat and SSR
**Describe the bug**
When an element is created in preact/compat with className="", rendering it to a string causes two class attributes to be output.

This can happen when using conditional classes returned from the npm `classnames` function, for example.

**To Reproduce**
```
const { createElement } = require("preact/compat");
const render = require("preact-render-to-string");
console.log(render(createElement("a", { className: "" })));
```
Tested with preact v10.6.4 and preact-render-to-string v5.1.19

**Observed behavior**
`<a class class></a>`

**Expected behavior**
`<a class></a>` or `<a class=""></a>`

rschristian:
Fixes #208

Empty `props.class` wouldn't be truthy, resulting in `class` being added to the element twice as outlined in the attached issue.

The added test might be a bit overkill, but it seems Mocha has no support for test isolation. Importing `preact/compat` (even in other suites) would throw off one of the existing tests, so I just run the compat suite subsequently. Let me know if there's a better way to go about this.

### DIFF
--- a/.changeset/fifty-rings-jog.md
+++ b/.changeset/fifty-rings-jog.md
@@ -1,0 +1,5 @@
+---
+'preact-render-to-string': patch
+---
+
+On empty className w/ compat, class attribute will no longer be duplicated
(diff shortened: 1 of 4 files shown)


JoviDeCroock (approved):

marvinhagemeister (approved):
Good catch!

PR #212, opened 2022-02-20
Support preact v11

JoviDeCroock:
We're rolling out our experimental v11 version, which switches us from storing everything on vnode/component to an internal


marvinhagemeister:
FYI: This needs to be marked as a breaking change.

marvinhagemeister (approved):

PR #224, opened 2022-06-16
fix: Serialize defaultChecked as checked attr

### LINKED ISSUE #223: Attribute `defaultChecked` should be outputted as `checked` instead of `defaultChecked`
**Describe the bug**
The `defaultChecked` boolean attribute is not properly outputted in the html.

**To Reproduce**

Run the following code in a Node REPL:

```javascript
const { createElement } = await import('preact')
const { renderToString } = await import('preact-render-to-string')

const html = renderToString(
  createElement('input', {
    type: 'checkbox',
    defaultChecked: true,
  }),
)

console.log(html)
// <input type="checkbox" defaultChecked />
```

**Expected behavior**

The output should be `<input type="checkbox" checked />`

rschristian:
Closes #223

Not sure if there are any other `defaultX` props to be concerned with (my memory and MDN are failing me). I think every other check / replace method would be a fair bit slower, so I imagine while limited to these two (`defaultValue` and `defaultChecked`) this approach is better? LMK!

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -263,10 +263,12 @@ function _renderToString(vnode, context, opts, inner, isSvgMode, selectValue) {
 
 			if (name === 'defaultValue') {
 				name = 'value';
+			} else if (name === 'defaultChecked') {
+				name = 'checked';
 			} else if (name === 'className') {
 				if (typeof props.class !== 'undefined') continue;
 				name = 'class';
-			} else if (isSvgMode && name.match(/^xlink:?./)) {
+			} else if (isSvgMode && /^xlink:?./.test(name)) {
 				name = name.toLowerCase().replace(/^xlink:?/, 'xlink:');
 			}
 
--- a/test/render.test.js
+++ b/test/render.test.js
@@ -95,6 +95,13 @@ describe('render', () => {
 			expect(rendered).to.equal(expected);
 		});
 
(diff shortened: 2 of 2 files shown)


src/index.js
```diff
@@ -263,10 +263,12 @@ function _renderToString(vnode, context, opts, inner, isSvgMode, selectValue) {
 
 			if (name === 'defaultValue') {
 				name = 'value';
+			} else if (name === 'defaultChecked') {
+				name = 'checked';
 			} else if (name === 'className') {
 				if (typeof props.class !== 'undefined') continue;
 				name = 'class';
-			} else if (isSvgMode && name.match(/^xlink:?./)) {
+			} else if (isSvgMode && /^xlink:?./.test(name)) {
```
rschristian, line 271: Unrelated (sorry!) but `.test()` should be marginally faster.

JoviDeCroock (approved):

alvinleung1996:
There is also a `defaultSelected` property for the `option` element.

https://html.spec.whatwg.org/multipage/form-elements.html#dom-option-defaultselected

JoviDeCroock:
@rschristian do you want this merged already? Wanted to do a small release for the perf improvements

rschristian:
@JoviDeCroock Sorry, must've missed or forgotten about this. Added in `defaultSelected`, so should probably be good unless you have suggestions or changes.

PR #225, opened 2022-07-05
chore: do several perf improvements

JoviDeCroock:
Uses a smarter stringbuilder props to @developit https://esbench.com/bench/5c7aaeef4cd7e6009ef6217c

### DIFF
--- a/.changeset/twelve-kiwis-tap.md
+++ b/.changeset/twelve-kiwis-tap.md
@@ -1,0 +1,5 @@
+---
+'preact-render-to-string': patch
+---
+
+Optimize perf by using smarter string builder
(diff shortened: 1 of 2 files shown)


developit:
Nice. It looks like this is currently faster on the `SearchResults` bench, but slower on the `Text` bench. My guess is that it's [this indexed access](https://github.com/preactjs/preact-render-to-string/pull/225/files#diff-bfe9874d239014961b1ae4e89875a6155667db834a410aaaa2ebe3cf89820556R385).

JoviDeCroock:
@developit restored that one

developit:
@JoviDeCroock I think we should consider dropping all of the pretty-printing from renderToString. It's only used by `render-to-string/jsx` afaik, and I don't know how widely-used that entry is anyway.

marvinhagemeister (approved):

PR #240, opened 2022-08-20
fix grid properties

### LINKED ISSUE #239: `px` unit for grid gap discrepancy between preact-render-to-string and react-dom/server #218
We've noticed a discrepancy in how preact / react server renders a component that looks like this:

```
import React from "preact/compat";
import "./styles.css";

export default function App() {
  return (
    <div
      style={{
        display: "grid",
        "grid-column-gap": 16,
        "grid-row-gap": 24,
        border: "1px solid red"
      }}
    >
      <div className="child">Child</div>
      <div className="child">Child</div>
      <div className="child">Child</div>
      <div className="child">Child</div>
    </div>
  );
}
```

Code sandboxes for preact and react to compare the output:

* [Preact example](https://codesandbox.io/s/preact-grid-gap-px-example-3g6uhn) 
* [React example](https://codesandbox.io/s/react-grid-gap-px-example-knevoe?file=/src/App.js:0-479)

We're using React 16.14.0 but as far as I can tell, React 17.x would have the same discrepancy.

I think the difference is that in Preact, we use this regex:

```
IS_NON_DIMENSIONAL = /acit|ex(?:s|g|n|p|$)|rph|grid|ows|mnc|ntw|ine[ch]|zoo|^ord|itera/i;
```

Which would match `grid-gap`, but React explicitly lists non dimensional properties: https://github.com/facebook/react/blob/v17.0.0/packages/react-dom/src/shared/CSSProperty.js#L11

So properties that would be non dimensional in Preact but not React are:

* gridRowGap
* gridColumnGap

There might be more, it might be worth auditing [gap](https://developer.mozilla.org/en-US/docs/Web/CSS/gap) properties fully to ensure there aren't more discrepancies

JoviDeCroock:
Fixes #239

### DIFF
--- a/.changeset/tender-pots-invent.md
+++ b/.changeset/tender-pots-invent.md
@@ -1,0 +1,5 @@
+---
+"preact-render-to-string": patch
+---
+
+fix grid properties in style attributes
--- a/src/util.js
+++ b/src/util.js
@@ -1,4 +1,5 @@
 // DOM properties that should NOT have "px" added when numeric
+// copied from Preact core to keep these consistent
 export const IS_NON_DIMENSIONAL = /^(-|f[lo].*[^se]$|g.{5,}[^ps]$|z|o[pr]|(W.{5})?[lL]i.*(t|mp)$|an|(bo|s).{4}Im|sca|m.{6}[ds]|ta|c.*[st]$|wido|ini)/;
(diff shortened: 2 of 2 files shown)


marvinhagemeister (approved):

JoviDeCroock:
Closing due to inactivity

PR #243, opened 2022-09-06
allow for setting checked to false

JoviDeCroock:
Relates to https://github.com/preactjs/signals/issues/74


JoviDeCroock:
This is the intended behavior

PR #251, opened 2022-10-14
Add HTML comments support

dirkjonker:
Preact doesn't really support HTML comments, but it is possible to (mis)use it to generate HTML comments anyway, by creating an element with a name that starts with `!--` and ends with `--`, making sure it does not have any attributes. For example, this is being used in Fresh (see denoland/fresh#838) to enable islands.
The closing tag of such a "comment" tag however generates invalid HTML, i.e.

```javascript
h("!--foo--", null, children)
```
will generate
```html
<!--foo--> ... </!--foo-->
```
The closing tag: `</!--` is invalid HTML and might make it hard for some clients to deal with. Chrome/Firefox have a long track record of magically fixing broken HTML and they turn this into:
```html
<!--!--foo---->
```
If you plan to programmatically read data from comments (which is a perfectly valid thing to do), you will get some unexpected results as your data will now be prefixed with `!--` and suffixed with `--`.

This change adds "support" for HTML comments by at least not generating invalid HTML.

I'm not 100% sure what kind of change this is, I would think a patch since it's quite the edge case and it fixes rendering invalid HTML in this edge case. But please do correct me if I'm wrong.

### DIFF
--- a/.changeset/late-zoos-swim.md
+++ b/.changeset/late-zoos-swim.md
@@ -1,0 +1,5 @@
+---
+'preact-render-to-string': patch
+---
+
+Prevent rendering invalid HTML for "comment" elements
--- a/src/index.js
+++ b/src/index.js
@@ -389,6 +389,10 @@ function _renderToString(vnode, context, isSvgMode, selectValue, parent) {
 		return startElement + ' />';
 	}
 
+	if (type.startsWith('!--')) {
+		return s + '<' + type + '>';
+	}
+
 	return s + '</' + type + '>';
 }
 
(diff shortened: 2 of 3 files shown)


marvinhagemeister (commented):
Awesome, thanks for making this PR! Given that we don't support rendering HTML comments in the core library, I'm a bit hesitant to make comments look like that here in `preact-render-to-string`.

Spoke with @developit earlier today and we came up with using `dangerouslySetInnerHTML` on a `Fragment` instead. That would make it obvious that this is something special and the `dangerouslySetInnerHTML` would allow us to skip modifying the string in any way.

Something like:

```js
<Fragment dangerouslySetInnerHTML={{__html: "<!-- this is a comment -->"}} />
```

dirkjonker:
That makes sense. I'm not using comments myself directly but Fresh uses it as a means to get islands working. Currently that doesn't give any real issues, only the w3 validator gives some warnings about garbage tags.

PR #252, opened 2022-10-15
Use named exports in README

marvinhagemeister:
Some `README` changes:

- Use named exports instead of default export. We might want to drop the default export at some point
- rest is prettier changes when saving the file


README.md
```diff
@@ -3,20 +3,18 @@
 [![NPM](http://img.shields.io/npm/v/preact-render-to-string.svg)](https://www.npmjs.com/package/preact-render-to-string)
 [![Build status](https://github.com/preactjs/preact-render-to-string/actions/workflows/ci.yml/badge.svg)](https://github.com/preactjs/preact-render-to-string/actions/workflows/ci.yml)
 
-Render JSX and [Preact] components to an HTML string.
+Render JSX and [Preact](https://github.com/developit/preact) components to an HTML string.
```
rschristian, line 6: May as well while we're at it:

```suggestion
Render JSX and [Preact](https://github.com/preactjs/preact) components to an HTML string.
```

rschristian (approved):

rschristian:
Good call!

PR #255, opened 2022-10-19
Update test scripts to allow watch usage

marvinhagemeister:
When working on `preact-render-to-string` I find myself manually building the test command for watch usage. This PR rewrites our test scripts so that we can pass `--watch` to both `test:mocha` and `test:mocha:compat` to enable watch mode.


rschristian (approved):
Ah good call.

Can't remember why I wrote it that way originally. Didn't want to switch between the two? Dunno.

PR #260, opened 2022-10-31
feat: use comments instead of element as marker

jacob-ebey:
Summary:
- use comments to denote placeholders to not break semantic HTML structures
- use a custom-element to handle moving into place
    - this remove the need for a mutation observer
    - seems like a cool use of a ce and I figured why not? should probably perf test it though
- adds onError for the underlying renderToChunks implementation
- adds renderToPipeableStream

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -11,7 +11,7 @@ function initPreactIslandElement() {
 
 			var s,
 				e,
-				c = document.createNodeIterator(document, NodeFilter.SHOW_COMMENT);
+				c = document.createNodeIterator(document, 128);
 			while (c.nextNode()) {
 				let n = c.referenceNode;
 				if (n.data == 'preact-island:' + i) s = n;
@@ -26,9 +26,8 @@ function initPreactIslandElement() {
 					e.parentNode.removeChild(p);
 					p = e.previousSibling;
 				}
-				while (d.firstChild) {
-					e.parentNode.insertBefore(d.firstChild, e);
-				}
+				while (d.firstChild) e.parentNode.insertBefore(d.firstChild, e);
+
 				d.parentNode.removeChild(d);
 			}
 		}
--- a/src/client.js
+++ b/src/client.js
@@ -3,13 +3,13 @@
 function initPreactIslandElement() {
 	class PreactIslandElement extends HTMLElement {
 		connectedCallback() {
-			if (!this.isConnected) return;
+			var d = this;
+			if (!d.isConnected) return;
 
 			let i = this.getAttribute('data-target');
 			if (!i) return;
 
-			var d = this,
-				s,
+			var s,
 				e,
 				c = document.createNodeIterator(document, NodeFilter.SHOW_COMMENT);
 			while (c.nextNode()) {
--- a/src/client.js
+++ b/src/client.js
@@ -10,16 +10,10 @@ function initPreactIslandElement() {
 
 			var d = this;
 			function f(el) {
-				var a = [];
-				for (var j = 0; j < el.childNodes.length; j++) {
-					var n = el.childNodes[j];
-					if (n.nodeType === 8) {
-						a.push(n);
-					} else {
-						a.push(...f(n));
-					}
-				}
-				return a;
+				let r = [];
+				let c = document.createNodeIterator(el, NodeFilter.SHOW_COMMENT);
+				while (c.nextNode()) r.push(c.referenceNode);
+				return r;
 			}
 			var s, e;
 			for (var n of f(document)) {
--- a/src/client.js
+++ b/src/client.js
@@ -1,52 +1,58 @@
 /* eslint-disable no-var, key-spacing, object-curly-spacing, prefer-arrow-callback, semi, keyword-spacing */
 
-/**
- * @param {number} c Total number of hydration islands
- */
-function initPreactIslands(c) {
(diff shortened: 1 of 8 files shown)


src/client.js
```diff
@@ -1,52 +1,58 @@
...
+function initPreactIslandElement() {
+	class PreactIslandElement extends HTMLElement {
+		connectedCallback() {
+			if (!this.isConnected) return;
+
+			let i = this.getAttribute('data-target');
+			if (!i) return;
+
+			var d = this;
+			function f(el) {
```
jacob-ebey, line 12: Is there a better way to get HTML comments than this?

jacob-ebey: Ok, this *should* be more performant now: https://github.com/preactjs/preact-render-to-string/pull/260/files#diff-627fe08d8bba6a4fa76e7c5ed36ef6f16d126733e65e19236b391de0d34f1fe0R14

Decided to go with a NodeIterator. TIL

src/client.js
```diff
@@ -1,52 +1,48 @@
...
+				while (p != s) {
+					if (!p || p == s) break;
+
+					e.parentNode.removeChild(p);
+					p = e.previousSibling;
 				}
+				while (d.firstChild) e.parentNode.insertBefore(d.firstChild, e);
+
+				d.parentNode.removeChild(d);
 			}
```
developit, line 32: Doing a deep walk here seems expensive. What about using a hidden element to denote the start of any island, followed by a comment sibling to denote its end?

```html
  <link id="preact-island-00">
  <h1>hello world</h1>
  this is some text
  <!--preact-island-00-->
```

```suggestion
			var n = 'preact-island-' + id,
				s = document.getElementById(n),
				p = s.parentNode,
				e = s,
				b;
			while (e) {
				b = e.nextSibling;
				p.removeChild(e);
				if (e.nodeType === 8 && e.data === n) break;
				e = b;
			}
			while (b = d.firstChild) p.insertBefore(b, e);
```

jacob-ebey: Breaks semantic HTML for lists and whatnot. Same reason to not use a CE for denoting the fallback.

marvinhagemeister: Yeah I think there is no way around using comments as we always risk breaking HTML or CSS selectors (`:nth-child`, etc)

src/client.js
jacob-ebey: If we are worried about perf we can switch to TreeWalker, but NodeIterator has better browser support if I remember right: 
image

jacob-ebey: More benchmarks if we want to switch:
Chome:
```
Name         Operations per second    Average time, ms
walker       1.0 x 10^5               0.01                ==============================>
iterator     5.0 x 10^4               0.02                ===============>
recursion    7.7 x 10^3               0.13                ==>
```

Safari:
```
Name         Operations per second    Average time, ms (main.ts, line 70)
walker       9.1 x 10^4               0.01                ==============================>
iterator     6.3 x 10^4               0.02                =====================>
recursion    3.6 x 10^3               0.28                =>
```

jacob-ebey: Here's the source for the benchmark:
```ts
import Benchmark from "micro-benchmark";

function nest(_parent: HTMLElement, content: string, n: number) {
  let _el = document.createElement("div");
  _el.innerText = content;
  _parent.appendChild(_el);

  if (--n) nest(_el, content, n);
  else {
    _el.appendChild(document.createComment("FIND_ME"));
  }
}

nest(document.body, "👻", 2000);

function findCommentRecursive(node: ChildNode, text: string): Node | null {
  if (node.nodeType === Node.COMMENT_NODE && node.nodeValue === text) {
    return node;
  }

  let child = node.firstChild;
  while (child) {
    const found = findCommentRecursive(child, text);
    if (found) return found;
    child = child.nextSibling;
  }

  return null;
}

function findCommentNodeIterator(node: ChildNode, text: string): Node | null {
  const iterator = document.createNodeIterator(node, NodeFilter.SHOW_COMMENT);
  let current = iterator.nextNode();
  while (current) {
    if (current.nodeValue === text) return current;
    current = iterator.nextNode();
  }
  return null;
}

function findCommentTreeWalker(node: ChildNode, text: string): Node | null {
  const walker = document.createTreeWalker(node, NodeFilter.SHOW_COMMENT);
  let current = walker.nextNode();
  while (current) {
    if (current.nodeValue === text) return current;
    current = walker.nextNode();
  }
  return null;
}

let result = Benchmark.suite({
  specs: [
    {
      name: "recursion",
      fn: () => {
        findCommentRecursive(document.body, "FIND_ME");
      },
    },
    {
      name: "iterator",
      fn: () => {
        findCommentNodeIterator(document.body, "FIND_ME");
      },
    },
    {
      name: "walker",
      fn: () => {
        findCommentTreeWalker(document.body, "FIND_ME");
      },
    },
  ],
});

let report = Benchmark.report(result);
console.log(report);
```

src/client.js
developit: Does this matter much if the elements are removed immediately?

jacob-ebey: Yes because when the fallback is displayed (potentially a long while) styles / semantics will be broken.

marvinhagemeister (approved):
LGTM, the more I think about it, the more I agree that the comment approach is the right one to take. Also good idea on using CE semantics instead of a `MutationObserver` 👍

PR #263, opened 2022-11-08
fix the regex breaking the tests

JoviDeCroock:
The glob pattern introduced in https://github.com/preactjs/preact-render-to-string/pull/257/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R35 broke the tests running correctly, this moves stuff to their own dir and tweaks test running


marvinhagemeister (approved):
Good catch 👍

PR #259, opened 2022-10-29
Streaming renderer

marvinhagemeister:
This PR that adds a streaming renderer. Main difference to `preact-ssr-prepass` is that it only awaits suspense boundaries and keeps everything else synchronous. This avoids a flooding the event loop with microtasks.

Tried a custom element approach to attach lazy hydrated subtrees, but the more I think about it the more I'm leaning towards using comments as folks sent me more and more use cases where introducing an arbitrary element breaks CSS styling (even with `display: contents`) or where the semantic HTML structure falls apart (`ul > li`).

Marking as draft until it's ready and we might need to make modifications to core too.

PR #241, opened 2022-08-22
Perf Improvements 3

developit:
This improves performance by another 5-10% compared to master (5.2.2):
Screen Shot 2022-08-21 at 6 36 54 PM

It also corrects the type definitions, reduces size by ~50% and removes pretty-printing entirely from the default bundle. It can still be accessed via `preact-render-to-string/jsx`.

**Because of the exports/options change, I'm marking this as a major release.**

PR #267, opened 2022-11-13
Streaming renderer: factor chunking out of main entrypoint and rebase on #241

developit:
This reimplements #259 atop #241, so that we can merge both and publish it as a single major.

With these changes, the streaming implementation imposes _almost_ no overhead on non-streaming rendering. Here's a comparison against #241 (`baseline` is #241, `current` is this PR):

Screen Shot 2022-11-13 at 10 28 44 PM


argyleink:
Want!

osdiab:
What is the status of this PR? Looking forward to streaming SSR so that frameworks like Next and Remix could work with preact.

developit:
@JoviDeCroock should we merge this?

JoviDeCroock (approved):

PR #269, opened 2022-11-16
Remove text VNode casting

developit:
This removes the text VNode casting from #257, as we're now handling this difference internally in `preact/debug` (preactjs/preact#3801). I'm only seeing a 1% performance improvement as a result of this change, but that should be a result of having already removed one of the per-text-vnode function calls in #241.


developit:
This can be merged now that preactjs/preact#3801 has been merged.

JoviDeCroock (approved):

marvinhagemeister (approved):

rschristian:
Looks like this landed as part of #286

PR #270, opened 2022-11-16
Unmount option hook call performance

developit:
This is a tiny performance improvement for unmount option hook.

Screenshot 2022-11-15 at 7 21 34 PM


JoviDeCroock (approved):

marvinhagemeister (approved):
Great catch! Let's add a changeset too 👍

PR #278, opened 2023-02-18
Perf improvements v2

JoviDeCroock:


marvinhagemeister (approved):
This looks great! Let's add a changeset too

src/util.js
```diff
@@ -64,14 +65,17 @@ export function styleObjToCss(s) {
...
-				str +
-				name +
-				':' +
-				val +
-				(typeof val === 'number' && IS_NON_DIMENSIONAL.test(prop) === false
-					? 'px;'
-					: ';');
+			let suffix = ';';
+			if (SUFFIX_CACHE[name]) {
+				suffix = 'px';
```
andrewiggins, line 70: does this need to be `'px;'` (note the semicolon after the `px`)?

src/util.js
```diff
@@ -64,14 +65,17 @@ export function styleObjToCss(s) {
...
-			str =
-				str +
-				name +
-				':' +
-				val +
-				(typeof val === 'number' && IS_NON_DIMENSIONAL.test(prop) === false
-					? 'px;'
-					: ';');
+			let suffix = ';';
+			if (SUFFIX_CACHE[name]) {
```
andrewiggins, line 69: Do we need to check if `typeof val === 'number'` here as well? If the first call to render is `<div style={{ padding: 8 }} />` but the second is `<div style={{ padding: '8px' }} />`, will it still add the `'px'`?

andrewiggins (commented):
@JoviDeCroock Sorry I'm a little late to the game 😅 I did have a couple of quick questions when you get a chance

PR #280, opened 2023-02-24
create demo for streaming render

JoviDeCroock:
This currently contains the bug that we correctly stream in but when the hydration continues we wipe the `Suspense` state with nothing and after 3 seconds replace it with the Counter content.

Another peculiar thing is that if we comment out the `entry-client.jsx` code that after the Counter resolves on the server the `loading...` boundary gets replaced with.. nothing

https://github.com/preactjs/preact-render-to-string/pull/259

### DIFF
--- a/demo/src/entry-client.jsx
+++ b/demo/src/entry-client.jsx
@@ -1,4 +1,4 @@
-import { hydrate } from 'preact';
-import { App } from './App';
+// import { hydrate } from 'preact';
+// import { App } from './App';
 
-hydrate(<App />, document);
+// hydrate(<App />, document);
--- a/demo/src/entry-client.jsx
+++ b/demo/src/entry-client.jsx
@@ -1,4 +1,5 @@
 import { hydrate } from 'preact';
+import { LocationProvider } from 'preact-iso';
 import { App } from './App';
 
 const config = { attributes: true, childList: true, subtree: true };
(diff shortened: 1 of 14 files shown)


src/lib/client.js
```diff
@@ -22,13 +23,18 @@ function initPreactIslandElement() {
...
 				while (p != s) {
 					if (!p || p == s) break;
-
 					e.parentNode.removeChild(p);
 					p = e.previousSibling;
 				}
-				while (d.firstChild) e.parentNode.insertBefore(d.firstChild, e);
 
-				d.parentNode.removeChild(d);
+				requestAnimationFrame(() => {
```
JoviDeCroock, line 30: This fixes a bug where the app server-side renders the whole `document`, we have to wait for the `d.children` to actually be populated which is an `afterFrame`

demo/src/entry-client.jsx
```diff
@@ -0,0 +1,4 @@
+// import { hydrate } from 'preact';
+// import { App } from './App';
+
+// hydrate(<App />, document);
```
JoviDeCroock, line 4: This is currently commented out, uncommenting this will make the `CommentNode` dissapear for the closing comment and hence kill hydration

JoviDeCroock:
@developit The wiping stopped happening when we supported comments in Preact, I used the open PR we have and then everything worked correctly.

PR #282, opened 2023-02-25
Remove trailing space for void elements

### LINKED ISSUE #279: Void elements have whitespace at closing
Preact: ``
React: ``

Not a critical issue but would be nice to match react's render to string output and it avoids having to do further minification.

JoviDeCroock:
Marked as major in case people rely on this for tests

Fixes #279

### DIFF
--- a/.changeset/gentle-trains-attack.md
+++ b/.changeset/gentle-trains-attack.md
@@ -2,4 +2,6 @@
 'preact-render-to-string': major
 ---
 
-Remove trailing space for void_elements
+Remove trailing space for void_elements, this could fail some test_assertions as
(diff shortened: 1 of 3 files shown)


andrewiggins:
What's the benefit here? Seems like it could be a really difficult change for people to pick up if lots of tests use this.

JoviDeCroock:
The being more in-line with the output React produces and it being a tad smaller resounded with me, however open to rebuttal there :D

developit (approved):

developit:
Not sure how this works with Changesets, but could we somehow put this into whatever the next major is, rather than creating a major for this change on its own?

JoviDeCroock:
@developit added it

andrewiggins (approved):
React compat seems useful. Let's do it in next major

PR #286, opened 2023-03-18
check whether style needs a suffix with a set and remove debug casting

JoviDeCroock:
In Preact >= 10.13.0 we fixed the debug casting, this removes this check as we're doing a major either way.

This also optimises style casting by using a Set, removing the need for the cache and thus saving some memory


src/index.js
```diff
@@ -145,15 +145,6 @@ function _renderToString(vnode, context, isSvgMode, selectValue, parent) {
 			rendered =
 				rendered +
 				_renderToString(child, context, isSvgMode, selectValue, parent);
-
-			if (
```
JoviDeCroock, line 149: We fixed this in 10.13.0, I honestly think we shouldn't put a preact/debug fix in our hot path

src/index.js
```diff
@@ -338,6 +329,8 @@ function _renderToString(vnode, context, isSvgMode, selectValue, parent) {
 	}
 
 	if (UNSAFE_NAME.test(type)) {
+		// this seems to performs a lot better than throwing
+		// return '<!-- -->';
```
JoviDeCroock, line 333: This could give us an additional 5% boost

test/render.test.js
```diff
@@ -1313,7 +1313,7 @@ describe('render', () => {
 				);
 			}
 
-			expect(render(<App />)).to.equal('<div><p>P481</p><p>P476951</p></div>');
+			expect(render(<App />)).to.equal('<div><p>P0-0</p><p>P0-1</p></div>');
```
JoviDeCroock, line 1316: use-id handling seems to be different in Preact 10.13

src/util.js
```diff
@@ -50,9 +49,45 @@ export let isLargeString = (s, length, ignoreLines) =>
 	String(s).indexOf('<') !== -1;
 
 const JS_TO_CSS = {};
-const SUFFIX_CACHE = {};
 
-const CSS_REGEX = /([A-Z])/g;
+const IS_NON_DIMENSIONAL = new Set([
```
marvinhagemeister, line 53: I like this change! A `Set` makes a lot of sense and we aren't as size constraint for RTS compared to core.

src/index.js
```diff
@@ -22,6 +22,7 @@ const assign = Object.assign;
 
 // Global state for the current render pass
 let beforeDiff, afterDiff, renderHook, ummountHook;
+const parent = h(Fragment, null);
```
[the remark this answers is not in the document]
JoviDeCroock: Hmm yes I'll revert this

src/index.js
marvinhagemeister: I really like that idea!

src/index.js
marvinhagemeister: I'm a bit out of the loop. Which issue in `preact/debug` did this section work around?

JoviDeCroock: This one https://github.com/preactjs/preact-render-to-string/pull/257

marvinhagemeister (approved):
Great work as always!

src/util.js
```diff
@@ -63,17 +98,15 @@ export function styleObjToCss(s) {
...
-			let isNumber = typeof val === 'number';
-			if (isNumber && SUFFIX_CACHE[name]) {
-				suffix = 'px;';
-			} else if (
-				isNumber &&
-				IS_NON_DIMENSIONAL.test(prop.toLowerCase()) === false
+			if (
+				typeof val === 'number' &&
+				// Exclude custom-attributes
+				!name.startsWith('--') &&
```
andrewiggins, line 107: Is this check doing the same thing as line 98, i.e. are they both trying to check for custom properties? Do we get any perf benefit by consolidating them?

andrewiggins (approved):
🎉 Super stoked by all this perf work on RTS!

PR #294, opened 2023-03-28
Bring back missing exports from 5.x

### LINKED ISSUE #290: CommonJS bundle broken for v6
Hi,

the commonjs version of v6 seems to be broken. There are no exported functions from the package.

```shell
markusw@lapi:~/source/xing/source/act-app (renovate/preact-render-to-string-6.x)$ node 
Welcome to Node.js v18.15.0.
Type ".help" for more information.
> require('preact-render-to-string/package.json').version
'6.0.0'
> require('preact-render-to-string')
undefined
> 
```

marvinhagemeister:
We removed a couple named exports in `6.0.0`, but it turns out that this clashes with a couple of libraries in the Preact ecosystem that we use. This PR brings back all the missing exports, with the exception of `shallowRender` which was moved to `preact-render-to-string/jsx` in `6.0.0`.

On top of that I added a quick script to verify that all of them work to prevent such issues in the future.

Fixes #290

### DIFF
--- a/.changeset/curvy-rings-invent.md
+++ b/.changeset/curvy-rings-invent.md
@@ -1,0 +1,5 @@
+---
+'preact-render-to-string': patch
+---
+
+Bring back exports from 5.x to make migration easier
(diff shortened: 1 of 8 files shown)


andrewiggins (approved):
Thanks!

JoviDeCroock (approved):

PR #296, opened 2023-03-30
feat: streaming rendering with Suspense boundaries as flush trigger

JoviDeCroock:
**Changes**

- Updates against current master
- Polyfills mask as we stopped setting that explicitly without useId invocations

**Preact**

I use a build of Preact with comments enabled for the demo https://github.com/preactjs/preact/pull/4128

**Notes**

While creating this demo I noticed a bug where when we suspend twice from a streaming boundary we end up in an infinite loop, also we can't unset `_parent` while streaming this would need to happen after everything 😩 

Solved it by using `_vnodeId` to have consistent ids for Suspense boundaries, keep checking for additions to the boundaries with `forkSuspended` and more fixes.

**Demo**

The demo basically forces a less than ideal scenario:

- The first boundary gets flushed which gives us the title and the loading placeholder - hydration pauses
- On the server we are resolving the data for the second boundary (10 pokemons)
- The second boundary gets flushed which gives us the list of 10 pokemons - hydration resumes
- On the client we now hit `useQuery` which is suspense enabled and throws as it doesn't have the pokemons - hydration pauses
- On the client the 10 pokemons get fetched - hydration finishes

Normally front-end data-fetching libraries will include the data in a `<script>` tag that gets flushed to the client so the data-restoration is synchronous. This demo however is meant to show off the resumed hydration.


src/lib/chunked.js
```diff
@@ -49,7 +49,13 @@ function handleError(error, vnode, renderChild) {
...
 	if (!vnode) return;
 
-	const id = vnode[MASK] + this.suspended.length;
+	let root = vnode;
+	while (root !== null && !root[MASK] && root[PARENT] !== null) {
+		root = root[PARENT];
+	}
+
+	const mask = root[MASK] || (root[MASK] = [0, 0]);
+	const id = 'P' + mask[0] + '-' + mask[1]++ + this.suspended.length;
```
JoviDeCroock, line 58: Had to change this logic as we aren't setting the mask anymore

src/stream.js
```diff
@@ -0,0 +1,44 @@
...
+			})
+				.then(() => {
+					controller.close();
+					allReady.resolve();
+				})
+				.catch((error) => {
+					controller.error(error);
+					allReady.reject(error);
+				});
+			// controller.enqueue(encoder.encode(shell));
```
marvinhagemeister, line 37: Nit: commented code

jsx.d.ts
```diff
@@ -10,10 +10,8 @@ interface Options {
...
 }
 
-export default function renderToStringPretty(
+export default function render(
 	vnode: VNode,
 	context?: any,
 	options?: Options
 ): string;
-
-export function shallowRender(vnode: VNode, context?: any): string;
```
marvinhagemeister, line 19: Was this an outdated export?

src/internal.d.ts
```diff
@@ -0,0 +1,38 @@
+// import { VNode } from 'preact';
```
marvinhagemeister, line 1: Nit: commented code

src/pretty.js
```diff
@@ -135,6 +133,8 @@ function _renderToStringPretty(
 				!nodeName.prototype ||
 				typeof nodeName.prototype.render !== 'function'
 			) {
+				// let cctx = getContext(nodeName, context);
```
marvinhagemeister, line 136: Curious why this can be commented out as it#s supposedly passed in 150 . Was it an error in the code?

src/stream-node.js
```diff
@@ -0,0 +1,60 @@
...
+			options.onAllReady && options.onAllReady();
+			stream.end();
+		})
+		.catch((error) => {
+			stream.destroy(error);
+		});
+
+	Promise.resolve().then(() => {
+		options.onShellReady && options.onShellReady();
+	});
```
marvinhagemeister, line 46: Would this fire when the rendering of the shell already errors? Not sure if the expected behavior is to call `onShellReady` regardless or only in the error case.

marvinhagemeister (approved):
This is great! Left some minor comments, but I'm good with landing this as early as possible and iterating on it in `main` 👍

src/pretty.js
JoviDeCroock: Code was duplicated with the branch above, fixed now

jsx.d.ts
JoviDeCroock: my bad, should be in there

src/stream-node.js
JoviDeCroock: The render runs synchronously, so after a micro-tick this just tells us that the synchronous shell render is done

marvinhagemeister (approved):

Issue #297, opened 2023-04-12
Unable to import shallowRender from preact-render-to-string/jsx due to TypeScript type errors


shinyama-k:
I am trying to import shallowRender using the following method, but I am encountering TypeScript type errors and cannot import it:

```
import { shallowRender } from 'preact-render-to-string/jsx';
```
I suspect that this is due to the missing declaration of `jsx.d.ts` in `package.json`.
https://github.com/preactjs/preact-render-to-string/blob/a9817ff0f177d58f5c8401bd3fce9d9c948a8eec/package.json#L50

PR #298, opened 2023-04-14
Fix to add type file for jsx.js

shinyama-k:
Fixes: https://github.com/preactjs/preact-render-to-string/issues/297

When I build the following code with the tsc command, I get an error message.
```
import { shallowRender } from 'preact-render-to-string/jsx';
import {h} from "preact";

const Sample = () => <div>sample</div>

shallowRender(<Sample />);
```
The error message is as follows

```
❯ tsc -p tsconfig.json --strict
src/index.tsx:1:31 - error TS7016: Could not find a declaration file for module 'preact-render-to-string/jsx'. '/Users/shinyama-k/preact-render/node_modules/preact-render-to-string/jsx.js' implicitly has an 'any' type.
  Try `npm i --save-dev @types/preact-render-to-string` if it exists or add a new declaration (.d.ts) file containing `declare module 'preact-render-to-string/jsx';`

1 import { shallowRender } from 'preact-render-to-string/jsx';

```

To fix this problem, added configuration to include the jsx.d.ts file in tarball of the npm package.

### DIFF
--- a/.changeset/mean-coins-move.md
+++ b/.changeset/mean-coins-move.md
@@ -1,0 +1,5 @@
+---
+'preact-render-to-string': patch
+---
+
+Fix to add type file for jsx.js
--- a/package.json
+++ b/package.json
@@ -51,6 +51,7 @@
 		"src",
 		"dist",
 		"jsx.js",
+		"jsx.d.ts",
 		"typings.json"
 	],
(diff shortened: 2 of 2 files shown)


rschristian (approved):
Good catch, thanks!

PR #300, opened 2023-06-06
Add support for linking to Deno projects

marvinhagemeister:
Deno uses ESM resolution internally which requires file extensions, similar proper ESM in node and browsers. These changes shouldn't have any effect on the distributed package as we flatten the code to a single file before publish.


marvinhagemeister:
Not going to  add a changeset, since it doesn't affect consumers of this package. It's only for folks working on `preact-render-to-string` directly.

JoviDeCroock (approved):

PR #301, opened 2023-06-06
Add support for rendering HTML comments

marvinhagemeister:
This PR adds support for rendering HTML comments via `preact-render-to-string`. Instead of adding a new vnode type this leverages `Fragments`. This has the benefit that branching cost is minimal as `Fragments` are the least frequently used components and they sorta represent "nothing" in terms of client code.

`<Fragment UNSTABLE_comment="foobar" />` renders `<!--foobar-->`. When the prop is present, all other children are ignored.

This solves an issue with fresh which currently renders invalid HTML comments. Browsers happily parse over that, but it's annoying if you're testing with HTML validators on CI. See https://github.com/denoland/fresh/issues/838

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -165,7 +165,7 @@ function _renderToString(vnode, context, isSvgMode, selectValue, parent) {
 	// Invoke rendering on Components
 	if (typeof type === 'function') {
 		if (type === Fragment) {
-			// Fragments are the least use components of core that's why
+			// Fragments are the least used components of core that's why
 			// branching here for comments has the least effect on perf.
 			if (props.comment) {
 				return '<!--' + encodeEntities(props.comment || '') + '-->';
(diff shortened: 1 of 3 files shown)


src/index.js
```diff
@@ -165,6 +165,12 @@ function _renderToString(vnode, context, isSvgMode, selectValue, parent) {
 	// Invoke rendering on Components
 	if (typeof type === 'function') {
 		if (type === Fragment) {
+			// Fragments are the least use components of core that's why
```
JoviDeCroock, line 168:
```suggestion
			// Fragments are the least used components of core that's why
```

JoviDeCroock (approved):

marvinhagemeister:
Marked the name as unstable as proposed in yesterday's call.

Issue #275, opened 2022-12-29
Meta charset not rendered as expected


aminnairi:
**Describe the bug**
When rendering a meta for encoding the page from the server using `render`, it fails to display the correct attribute casing which should be lowercased for the `S` letter instead it is uppercased just like it is written in JSX.

**To Reproduce**
Follow instructions for the reproduction repository [here](https://github.com/aminnairi/preact-charset-issue-template).

**Expected behavior**
It should render the attribute `charSet` as `charset` when passing through the `render` function for the `<meta>` tag.

**Screenshots**
None.

**Browser (please complete the following information):**
 - OS: [e.g. Windows 11]: Arch
 - Browser [e.g. Chrome 108]: Chrome 108

**Node (please complete the following information):**
 - Platform: [e.g. Node 18]: Node 19.3.0
 - Package Manager [e.g. NPM 8.2.13]: NPM 8.19.2
 - Package Version [e.g. 0.2.1]: 5.2.6

**Additional context**
None.

JoviDeCroock:
Other frameworks don't do this either, it's `charset` and not `char-set` which the camelCased variant would imply, I quickly checked how [react goes about this](https://codesandbox.io/s/adoring-morning-6vpgfb?file=/src/App.js) and they handle it the same way, they don't touch it.

Issue #26, opened 2017-02-17
SVG attributes format


nrandell:
Not sure if this an issue or a clarification.
If I have some SVG, such as
`<path d="M 2 50 A 48 48 0 0 1 50 2" stroke="#gradLeft-1" strokeWidth="4" fill="none" />`
It doesn't work correctly.
If I change 'strokeWidth' to 'stroke-width' (changed to lower case with a dash) then it does work ok.
This is different to react. 
Is this behaviour correct? or have I done something wrong?

developit:
Ah yeah this looks like an issue.  In preact it's more common to use the attribute names, but we do officially support both.  We'll need to convert them :)

nrandell:
Great - good to know I wasn't being stupid!

Born2Go:
hello everybody, any progress in this issue? I just stumbled in the same problem

hufan-akari:
maybe the same issue.
current version  support `http-equiv` but not `httpEquiv`.
support  `charset ` but not `charSet`.

SukkaW:
> Ah yeah this looks like an issue. In preact it's more common to use the attribute names, but we do officially support both. We'll need to convert them :)

`react-dom` has a file `possibleStandardNames.js` that can be used to normalize attributes:

https://github.com/facebook/react/blob/ca106a02d1648f4f0048b07c6b88f69aac175d3c/packages/react-dom/src/shared/possibleStandardNames.js

cc @developit

PR #307, opened 2023-07-26
Attribute hook render option

gpoitch:
Adds an attribute hook rendering option. This allows the user to customize how attribute name casing is rendered.
There have been several issues and attempts over years to implement this. Namely: #199

Seems the consensus is to do nothing because other frameworks don't and there isn't an agreeable set of regexes to handle everything, so instead let the user do it.

From my experience, the proper casing on some of these is absolutely critical. We've been maintaining a fork for years, and couldn't keep up syncing upstream improvements after v6 so this is a more non-invasive approach.

### DIFF
--- a/jsx.d.ts
+++ b/jsx.d.ts
@@ -8,6 +8,7 @@ interface Options {
 	functions?: boolean;
 	functionNames?: boolean;
 	skipFalseAttributes?: boolean;
+	attributeHook?: (name: string) => string;
 }
 
 export default function renderToStringPretty(
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -1,27 +1,25 @@
 import { VNode } from 'preact';
 
-interface RenderOptions {
-	attrHook?: (name: string) => string;
+interface Options {
+	attributeHook?: (name: string) => string;
 }
 
 export default function renderToString(
 	vnode: VNode,
 	context?: any,
-	renderOpts?: RenderOptions
+	options?: Options
 ): string;
 
-export function render(
-	vnode: VNode,
-	context?: any,
-	renderOpts?: RenderOptions
-): string;
+export function render(vnode: VNode, context?: any, options?: Options): string;
+
 export function renderToString(
 	vnode: VNode,
 	context?: any,
-	renderOpts?: RenderOptions
+	options?: Options
 ): string;
+
 export function renderToStaticMarkup(
 	vnode: VNode,
 	context?: any,
-	renderOpts?: RenderOptions
+	options?: Options
 ): string;
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -1,7 +1,27 @@
 import { VNode } from 'preact';
 
-export default function renderToString(vnode: VNode, context?: any): string;
+interface RenderOptions {
+	attrHook?: (name: string) => string;
+}
 
-export function render(vnode: VNode, context?: any): string;
-export function renderToString(vnode: VNode, context?: any): string;
-export function renderToStaticMarkup(vnode: VNode, context?: any): string;
+export default function renderToString(
+	vnode: VNode,
+	context?: any,
+	renderOpts?: RenderOptions
+): string;
+
+export function render(
+	vnode: VNode,
+	context?: any,
+	renderOpts?: RenderOptions
+): string;
+export function renderToString(
+	vnode: VNode,
+	context?: any,
+	renderOpts?: RenderOptions
+): string;
+export function renderToStaticMarkup(
+	vnode: VNode,
+	context?: any,
+	renderOpts?: RenderOptions
+): string;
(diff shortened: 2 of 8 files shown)


marvinhagemeister:
Agree, to me it feels like that use case is already covered by `options.vnode`. Benefit of using that is that it also works in the browser if it's included in a script tag somewhere.

gpoitch:
It doesn't really work well.
- You have to reimplement everything here: https://github.com/preactjs/preact-render-to-string/blob/main/src/index.js#L323-L399
- Having trouble getting the hook to invoke at all in some cases

gpoitch:
Broken down to a simple example, there is no info to work with in the vnode: https://codesandbox.io/s/happy-payne-vz84fx?file=/src/index.js

marvinhagemeister:
@gpoitch Your codesandbox has a couple of issues and unnecessary cloning. The `options` object is a global and can be thought of as global event listeners. By attaching a new `options.vnode` hook on every render you're growing the chain of hooks by one with each render. There is no need to try to uppercase the same `props` object over and over again. Doing it once is enough.

I'm not sure why the `options` object is cloned to `defaultOptions`. Maybe that's a leftover from an earlier exploration? The `.reduce()` call can be simplified by using a `for-in` loop. The wrapper around `renderToString` doesn't do anything either.

Here is a cleaned up version of the codesandbox: https://codesandbox.io/s/intelligent-heisenberg-7h6ysl?file=/src/index.js

> You have to reimplement everything here: [main/src/index.js#L323-L399](https://github.com/preactjs/preact-render-to-string/blob/main/src/index.js?rgh-link-date=2023-07-26T16%3A03%3A29Z#L323-L399)

I guess the more interesting question we should ask here is: Why do you want to uppercase the attribute names in the first place? I think if you can present a case where this is needed we're more than happy to add it. Can you share more about that?

gpoitch:
@marvinhagemeister the uppercasing was just to keep the example simple.
The wrapper was for hopefully for installing different hooks when rendering different content, which doesn't appear possible with the options modifications.

Real use-cases for this of the top of my head:
- RSS. Doesn't validate/work in some readers without proper casing
- AMP validation fails
- srcSet breaks in certain browsers and downloads the fallback

marvinhagemeister:
> - RSS. Doesn't validate/work in some readers without proper casing

That's a fair point. Are those readers common?

> - SVGs not rendering correctly

Do you have an example of an SVG that renders incorrectly?

> - AMP validation fails

Do you have an example where AMP validation fails?

> - srcSet issues in certain browsers downloading wrong/excessive images

To me that sounds like a bug we should fix.

gpoitch:
> Do you have an example

Yes a media company that runs on this with 10s of millions of daily views 😄. I can put together examples if you really want but there is an entire class of content rendering that may not have preact/the browser "correcting" it.

I'd happily use the vnode option if it's viable but the already handled special cases and option re-modification doesn't look promising. We will continue to run the fork, just thought I'd share the idea since there several issues about this always popping up (`#26, #275, #18, #55`). Thanks for taking the time to look at this.

marvinhagemeister:
>> Do you have an example
>
> Yes a media company that runs on this with 10s of millions of daily views 😄. I can put together examples if you really want > but there is an entire class of content rendering that may not have preact/the browser "correcting" it.

It's great to hear that you have successful clients. I was asking for a technical example. I don't work frequently with RSS readers so my question was more about if that's spec thing in that it only allows uppercase attribute names or if the spec is case insensitive and the readers implemented it incorrectly.

Regarding #26, #275, #18, #55: I'd rather fix those issues in `preact-render-to-string` itself rather than giving users a hook to work around these. That way everyone can profit from the same fixes. What do you think?

gpoitch:
Nothing uses uppercase. Again, that was for a simple example's sake.  All of the real transformations I've encountered [are in the tests](https://github.com/preactjs/preact-render-to-string/pull/307/files#diff-44825949d238c8f1fb773ea0b52e753fdd59831e5e56691414a14be8a24ef481R1604-R1615) and #199 

The technical examples are literally every single RSS, SVG, and AMP page you write in JSX and render with this library. If the RSS doesn't fully validate, clients/apps won't accept your content. If the AMP doesn't validate, google won't put you in search. If the SVG doesn't have proper attributes it won't render correctly.

Regarding the other issues and correcting the attributes directly, I already provide that in #199 and got no traction. We could do a combination of the 2 - provide this hook and either export or make [this](https://github.com/preactjs/preact-render-to-string/pull/307/files#diff-44825949d238c8f1fb773ea0b52e753fdd59831e5e56691414a14be8a24ef481R1604-R1615) the default. If you have another direction in mind, I will gladly implement it.

marvinhagemeister:
Just merged #308 which addresses the known HTML + SVG scenarios. I'm not too familiar with AMP or RSS. Do they have additional cases to be aware of?

gpoitch:
For RSS: `xmlnsFoo` -> `xmlns:foo` 
`/^(xlink|xml|xmlns)([A-Z])/` is the regex I was using to colon-ize
(real attrs: xmlns:dc, xmlns:content, xmlns:atom, xmlns:media) 

AMP should be handled with the HTML scenarios

PR #313, opened 2023-08-13
fix: escape a single quote

yusukebe:
Hi,

Firstly, thank you for the great project.

In this PR, I've implemented the escaping of a single quote (0x27) to `&#x27;`. This modification will prevent the potential execution of scripts, as illustrated below:

```tsx
const value = "alert('bar!')";
return <div onMouseOver={value}>foo</div>;
```


marvinhagemeister:
FYI: This is a breaking change. A a good chunk of users from the [Fresh framework](https://fresh.deno.dev/) depend on this working.

yusukebe:
Hi @marvinhagemeister,

I'm aware that Preact is used for Fresh, and I a fan of it. Indeed, this change introduces a breaking change that could have a significant impact. I believe it would be best to include this change when this package is released with a major version upgrade.

.changeset/quiet-dragons-lay.md
```diff
@@ -0,0 +1,5 @@
+---
+"preact-render-to-string": patch
```
rschristian, line 2:
```suggestion
"preact-render-to-string": major
```

Issue #328, opened 2024-01-11
Incorrect jsx type definition


acelaya:
We have just started to use `preact-render-to-string/jsx` and noticed TypeScript complains because [`jsx.d.ts`](https://github.com/preactjs/preact-render-to-string/blob/cb32a6a4e0b3278ec343a66b1bf8f860309741cd/jsx.d.ts) includes two default exports.

As a temporary workaround we have created our own definition and referenced it via tsconfig's [`paths`](https://www.typescriptlang.org/tsconfig#paths) option.

We could provide a fix for this if desired, consisting in removing the last `export default render;`, which looks like an overlook.

marvinhagemeister:
Good catch, feel free to send a PR. The last default export is definitely a mistake and shouldn't be there.

acelaya:
Thanks! Here's the PR https://github.com/preactjs/preact-render-to-string/pull/329

PR #329, opened 2024-01-11
Remove incorrect second default export from jsx.d.ts

acelaya:
This PR fixes https://github.com/preactjs/preact-render-to-string/issues/328, by removing the second default export from `jsx.d.ts`

### DIFF
--- a/.changeset/early-rules-nail.md
+++ b/.changeset/early-rules-nail.md
@@ -1,0 +1,5 @@
+---
+'preact-render-to-string': patch
+---
+
+Remove incorrect second default export from jsx.d.ts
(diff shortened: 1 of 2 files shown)


JoviDeCroock (approved):
Just needs a changeset and we're good to go

acelaya:
> Just needs a changeset and we're good to go

Just pushed it.

Not related with this PR, but I had to change `"baseBranch": "master"` to `"baseBranch": "main"` in changeset's config file, in order to make changeset cli work.

I then reverted it to avoid introducing unrelated changes here, but just FYI.

PR #331, opened 2024-01-13
fix: shallow render with fragment at top lvl

### LINKED ISSUE #262: Rendering shallow makes a Fragment into a <p>
Shallow rendering this:
`<><ComponentX /></>`
Creates this:
`<p><ComponentX /></p>`

Is that intended behaviour? Why not just `<ComponentX />`

immitsu:
Fixes #262

### DIFF
--- a/src/pretty.js
+++ b/src/pretty.js
@@ -103,10 +103,12 @@ function _renderToStringPretty(
 	// components
 	if (typeof nodeName === 'function') {
 		isComponent = true;
-		if (opts.shallow && (inner || opts.renderRootComponent === false)) {
-			nodeName = getComponentName(nodeName);
-		} else if (nodeName === Fragment) {
-			const children = [];
+
+		let isFragment = nodeName === Fragment,
+			isShallow = opts.shallow && (inner || opts.renderRootComponent === false);
+
+		if (isFragment || (isShallow && isFragment)) {
+			let children = [];
 			getChildren(children, vnode.props.children);
 			return _renderToStringPretty(
 				children,
@@ -116,6 +118,8 @@ function _renderToStringPretty(
 				isSvgMode,
 				selectValue
 			);
+		} else if (isShallow) {
+			nodeName = getComponentName(nodeName);
 		} else {
 			let rendered;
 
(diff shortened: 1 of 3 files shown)


remko79:
Can we get this reviewed and merged please?

src/pretty.js
```diff
@@ -104,10 +104,12 @@ function _renderToStringPretty(
...
 		isComponent = true;
-		if (opts.shallow && (inner || opts.renderRootComponent === false)) {
-			nodeName = getComponentName(nodeName);
-		} else if (nodeName === Fragment) {
-			const children = [];
+
+		let isFragment = nodeName === Fragment,
+			isShallow = opts.shallow && (inner || opts.renderRootComponent === false);
+
+		if (isFragment || (isShallow && isFragment)) {
```
rschristian, line 111: This conditional doesn't make much sense

src/pretty.js
```diff
@@ -104,10 +104,12 @@ function _renderToStringPretty(
...
+			let children = [];
```
rschristian, line 112: nit: let's keep it as `const`, reduces the diff

```suggestion
			const children = [];
```

src/pretty.js
```diff
@@ -104,10 +104,12 @@ function _renderToStringPretty(
 	// components
 	if (typeof nodeName === 'function') {
 		isComponent = true;
-		if (opts.shallow && (inner || opts.renderRootComponent === false)) {
-			nodeName = getComponentName(nodeName);
-		} else if (nodeName === Fragment) {
-			const children = [];
+
+		let isFragment = nodeName === Fragment,
+			isShallow = opts.shallow && (inner || opts.renderRootComponent === false);
```
rschristian, line 109: Can put these right back in the `if()` statements

remko79:
Not sure if I should put this here, but else I can move it to my issue. I think we should only add the check ` && nodeName !== Fragment` on line 107 and omit other changes (except the updated test). The full diff based on current main branch is attached here.
[diff.txt](https://github.com/preactjs/preact-render-to-string/files/15250257/diff.txt)

rschristian (approved):
Thanks! Sorry for the delay.

PR #303, opened 2023-06-09
Simple Suspense renderer

lemonmade:

PR #333, opened 2024-02-08
Simple suspense renderer 2024

JoviDeCroock:
Supersedes #303


test/compat/async.test.js
```diff
@@ -0,0 +1,68 @@
+import render from '../../src/index.js';
+import {
+	h,
+	Component,
+	createContext,
+	Fragment,
+	options,
+	createRef
+} from 'preact';
```
JoviDeCroock, line 9:
```suggestion
import { h } from 'preact';
```

marvinhagemeister (approved):

src/index.js
```diff
@@ -60,8 +60,74 @@ export function renderToString(vnode, context) {
...
 			false,
 			undefined,
-			parent
+			parent,
+			false
 		);
+	} catch (e) {
+		if (e.then) {
+			throw new Error('Use "renderToStringAsync" for suspenseful rendering.');
+		}
```
rschristian, line 69: Only just noticed, this is breaking for all preact-iso users... do we want to downgrade this to a warning message instead of throwing an error, or just accept the breaking change in a minor?

PR #335, opened 2024-02-19
Fixing rendering multiple suspended components within one Suspense boundary

dios-david:
Rendering a Suspense boundary which have multiple suspended components as children fails:

```ts
const promise = renderToStringAsync(
  <ul>
    <Suspense fallback={null}>
      <SuspenderOne>
        <li>one</li>
      </SuspenderOne>
      <SuspenderTwo>
        <li>two</li>
      </SuspenderTwo>
    </Suspense>
  </ul>
);
```

```
  1) Async renderToString
       should render JSX with multiple suspended direct children within a single suspense boundary:
     Error: the promise {} was thrown, throw an Error :)
```

I was curious so I trying to debug and fix this, what I found is:
- `_renderToString` (with `asyncMode: true`) throws a Promise while rendering the children of a component, if any of the children is a suspended component
- in this case the error Promise is caught, and once it resolves the same render is called again (without the try-catch this time)
- but when that second render throws again (e.g. when another suspended child component is still not resolved), that Promise is not caught

I changed this a bit by moving the _"render children, if that throws a Promise then render again once that Promise is resolved"_ bit to a function and re-using that whenever rendering a children. This way if a Suspense component has multiple suspended child components, we keep rendering until all of the promises are resolved.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -427,9 +427,8 @@ function _renderToString(
 			rendered != null && rendered.type === Fragment && rendered.key == null;
 		rendered = isTopLevelFragment ? rendered.props.children : rendered;
 
-		try {
-			// Recurse into children before invoking the after-diff hook
-			const str = _renderToString(
+		const renderChildren = () =>
+			_renderToString(
 				rendered,
 				context,
 				isSvgMode,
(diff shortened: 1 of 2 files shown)


JoviDeCroock (approved):

marvinhagemeister (approved):

src/index.js
```diff
@@ -438,16 +438,20 @@ function _renderToString(
...
 				context,
 				isSvgMode,
 				selectValue,
 				vnode,
 				asyncMode
 			);
+
+		try {
+			// Recurse into children before invoking the after-diff hook
+			const str = renderChildren();
```
marvinhagemeister, line 453: Worth benchmarking in the future. Creating inline closures is likely slower than inlining it into the various places.

PR #352, opened 2024-05-11
fix: Ensure `cellPadding`, `cellSpacing`, and `useMap` are lower-cased

rschristian:


JoviDeCroock (approved):
Did we encounter an issue somewhere for this?

rschristian:
I'm not aware of any issue, no. I was fiddling around with the `popoverTarget` PR and did a quick check of lower casing all names just to see what stood out.

I'm guessing #311 didn't want to handle this (reasonable, given the scope of what they were doing) and it's not intentional? Especially as these 3 do have valid attributes that we should be preferring.

src/lib/util.js
```diff
@@ -1,7 +1,7 @@
 export const VOID_ELEMENTS = /^(?:area|base|br|col|embed|hr|img|input|link|meta|param|source|track|wbr)$/;
 export const UNSAFE_NAME = /[\s\n\\/='"\0<>]/;
 export const NAMESPACE_REPLACE_REGEX = /^(xlink|xmlns|xml)([A-Z])/;
-export const HTML_LOWER_CASE = /^accessK|^auto[A-Z]|^ch|^col|cont|cross|dateT|encT|form[A-Z]|frame|hrefL|inputM|maxL|minL|noV|playsI|popoverT|readO|rowS|spellC|src[A-Z]|tabI|item[A-Z]/;
+export const HTML_LOWER_CASE = /^accessK|^auto[A-Z]|^cell|^ch|^col|cont|cross|dateT|encT|form[A-Z]|frame|hrefL|inputM|maxL|minL|noV|playsI|popoverT|readO|rowS|spellC|src[A-Z]|tabI|useM|item[A-Z]/
```
rschristian, line 4:
```suggestion
export const HTML_LOWER_CASE = /^accessK|^auto[A-Z]|^cell|^ch|^col|cont|cross|dateT|encT|form[A-Z]|frame|hrefL|inputM|maxL|minL|noV|playsI|popoverT|readO|rowS|spellC|src[A-Z]|tabI|useM|item[A-Z]/;
```

PR #362, opened 2024-05-24
fix: Incorrect entries in stream exports

rschristian:


config/node-13-exports.js
```diff
@@ -1,11 +1,7 @@
 const fs = require('fs');
 
-const snakeCaseToCamelCase = str =>
-	str.replace(/([-_][a-z])/g, group => group.toUpperCase().replace('-', ''));
```
rschristian, line 4: I don't know what this was for, but we certainly don't need it

package.json
```diff
@@ -25,14 +25,14 @@
...
 			"types": "./stream.d.ts",
-			"import": "./dist/stream.mjs",
 			"browser": "./dist/stream.module.js",
+			"import": "./dist/stream.mjs",
 			"require": "./dist/stream.js"
 		},
 		"./stream-node": {
 			"types": "./stream-node.d.ts",
-			"import": "./dist/stream-node.mjs",
 			"browser": "./dist/stream-node.module.js",
```
rschristian, line 34: I think this should be removed? It's not suitable for browser use, it imports from `node:stream`.

package.json
```diff
@@ -24,15 +24,14 @@
...
+			"types": "./dist/stream.d.ts",
 			"browser": "./dist/stream.module.js",
+			"import": "./dist/stream.mjs",
 			"require": "./dist/stream.js"
 		},
 		"./stream-node": {
-			"types": "./stream-node.d.ts",
+			"types": "./dist/stream-node.d.ts",
 			"import": "./dist/stream-node.mjs",
-			"browser": "./dist/stream-node.module.js",
```
rschristian, line 35: Dropped as it's not suitable for browser use -- it imports from `node:stream`.

JoviDeCroock (approved):

PR #363, opened 2024-05-24
refactor: Build scripts & output directories

rschristian:
Builds on top of #362-ish

Does a few things:

- Output in `dist/` is now nested per export
  - Makes it so much easier to debug/deal with when the exports aren't all at same level.
- Simplifies & makes build scripts a tad more consistent
- Drops `pretty-format` dependency -- it's already inlined into the export that uses it. It's not a heavy dep, but entirely unused all the same.

### DIFF
--- a/config/node-commonjs.js
+++ b/config/node-commonjs.js
@@ -8,7 +8,6 @@ const filePath = (file) => path.join(process.cwd(), 'dist', file);
 
 // Main entry
 fs.copyFileSync(filePath('index.js'), filePath('commonjs.js'));
-fs.copyFileSync(filePath('index.js.map'), filePath('commonjs.js.map'));
 
 const source = [
 	`const mod = require('./commonjs');`,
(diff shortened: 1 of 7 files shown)


config/node-commonjs.js
```diff
@@ -8,7 +8,6 @@ const filePath = (file) => path.join(process.cwd(), 'dist', file);
 
 // Main entry
 fs.copyFileSync(filePath('index.js'), filePath('commonjs.js'));
-fs.copyFileSync(filePath('index.js.map'), filePath('commonjs.js.map'));
```
rschristian, line 11: No point in copying the maps AFAIK, we don't update the source map comment, i.e., `commonjs.js` still has `//# sourceMappingURL=index.js.map`.

Could fix that,  but I don't see a real need to?

JoviDeCroock (approved):

PR #367, opened 2024-05-26
fix: Add types for renderStream & pipeableStream

rschristian:
Adds hand-written types for `stream` & `node-stream` exports. While Microbundle can generate types, it seems the types it generates for the main & `jsx` entries are a bit wonky and don't get copied over correctly, so I chose to simply match the existing practice and write some definitions myself. They're simple enough anyways.

I also added `check-export-map`, as we have in core, to catch this sort of issue in the future.


jsx.d.ts
```diff
@@ -10,10 +10,15 @@ interface Options {
...
 	options?: Options
 ): string;
+export function render(vnode: VNode, context?: any, options?: Options): string;
 
-export function shallowRender(vnode: VNode, context?: any): string;
+export function shallowRender(
+	vnode: VNode,
+	context?: any,
+	options?: Options
+): string;
```
rschristian, line 24: Changes here are just to keep it in sync with `src/jsx.d.ts`

package.json
```diff
@@ -39,13 +39,12 @@
...
-		"build": "npm run -s transpile && npm run -s transpile:jsx && npm run -s transpile:stream && npm run -s transpile:stream-node && npm run -s copy-typescript-definition",
-		"postbuild": "node ./config/node-13-exports.js && node ./config/node-commonjs.js && node ./config/node-verify-exports.js",
+		"build": "npm run -s transpile && npm run -s transpile:jsx && npm run -s transpile:stream && npm run -s transpile:stream-node",
+		"postbuild": "node ./config/node-13-exports.js && node ./config/node-commonjs.js && node ./config/node-verify-exports.js && check-export-map",
 		"transpile": "microbundle src/index.js -f es,cjs,umd",
 		"transpile:stream": "microbundle src/stream.js -o dist/stream/index.js -f es,cjs,umd",
 		"transpile:stream-node": "microbundle src/stream-node.js -o dist/stream/node/index.js -f es,cjs,umd --target node",
 		"transpile:jsx": "microbundle src/jsx.js -o dist/jsx/index.js -f es,cjs,umd && microbundle dist/jsx/index.js -o dist/jsx/index.js -f cjs",
-		"copy-typescript-definition": "copyfiles -f src/*.d.ts dist",
```
rschristian, line 48: Won't need anymore, Microbundle will copy the type definitions over itself.

rschristian: Ope, I'm wrong, was looking at a non-clean build of `dist/`.

JoviDeCroock (approved):

marvinhagemeister (approved):

PR #368, opened 2024-05-29
chore: use `.jsx` extension for files with JSX

marvinhagemeister:


rschristian (approved):
Looking into other test tooling?

marvinhagemeister:
Nah, just using this repo as a test bed for node compatibility stuff in Deno

PR #372, opened 2024-06-13
fix: insert ooo chunks in the proper order

jacob-ebey:
fix: stop client runtime from being corrupted

This does two things
1.) the current logic to move OOO children doesn't support multiple children, and is kinda just wrong
2.) puts the client OOO runtime into a string to stop corruption by bundlers such as Vite

### DIFF
--- a/.changeset/quick-goats-end.md
+++ b/.changeset/quick-goats-end.md
@@ -1,0 +1,6 @@
+---
+"preact-render-to-string": patch
+---
+
+fix: stop client runtime from being corrupted
+fix: insert ooo chunks in the proper order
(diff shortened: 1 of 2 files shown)


JoviDeCroock (approved):
Great spot, thank you! If you have the time mind describing in what way this would fail with multiple children? As we don't have tests it might be good to have some shared knowledge here.

jacob-ebey:
> Great spot, thank you! If you have the time mind describing in what way this would fail with multiple children? As we don't have tests it might be good to have some shared knowledge here.

It is currently looping over an array and incrementing the index that it's mutating.
Start with: [0, 1]
Loop 1:
i = 0
get and remove at i ending with [1]
Loop 2:
i = 1
can't get and remove anything at 1 as what was there is now at 0.

The tried and true "iterate until you have no more children" is best in this case. And we just `.after()` on the previously moved node starting at the island comment.

PR #374, opened 2024-06-13
fix: hydration might remove the node from the DOM causing it no longer have a parent

jacob-ebey:


jacob-ebey:
@JoviDeCroock Got another one for ya.

jacob-ebey:
Nevermind, this isn't the core of the issue I'm seeing. Need support for preserving the comments during hydration. There are race conditions there.

JoviDeCroock:
@jacob-ebey comments should be preserved with https://github.com/preactjs/preact/pull/3327 - do you think Preact is removing it or is there something else going on?

Issue #377, opened 2024-07-15
renderToStringAsync can be confused by a Promise from _renderToString 


ccouzens:
Hey

I'm getting an issue in my application where `await renderToStringAsync` returns something along the lines of

```
Promise(['html', Promise(['html', 'html'])])
```

The issue is that in my app `_renderToString` [here](https://github.com/preactjs/preact-render-to-string/blob/main/src/index.js#L112-L120) returns a Promise that resolves into arrays (of further promises), rather than an array that contains promises.

I've fixed this locally for myself by making this addition.

```
		if (rendered instanceof Promise) {
			rendered = [rendered];
		}
```

inside `renderToStringAsync` before checking if it's an array.

I would raise this as a PR, but I'm currently unable to reproduce it in a test case.

Hopefully tomorrow I'll extract a test case from my app and raise a PR. In the meantime, do you know what would cause `_renderToString` to return a Promise of arrays, rather than an array of promises?

Here's my WIP https://github.com/preactjs/preact-render-to-string/compare/main...ccouzens:preact-render-to-string:patch-1?expand=1

Thanks,
Chris

PR #378, opened 2024-07-17
Fix instance where renderToStringAsync returns a promise of a promise

ccouzens:
Fixes https://github.com/preactjs/preact-render-to-string/issues/377

I have had this happen with a project I work on. This includes a minimal test reproduction of it.

### DIFF
--- a/test/compat/async.test.jsx
+++ b/test/compat/async.test.jsx
@@ -1,5 +1,5 @@
 import { renderToStringAsync } from '../../src/index.js';
-import { h, createContext, Fragment } from 'preact';
+import { h, Fragment } from 'preact';
 import { Suspense, useId, lazy } from 'preact/compat';
 import { expect } from 'chai';
 import { createSuspender } from '../utils.jsx';
@@ -209,21 +209,15 @@ describe('Async renderToString', () => {
 			return <Fragment>{children}</Fragment>;
 		};
 
-		const ThemeContext = createContext('light');
-
 		const LazyComponent = lazy(async () => {
-			await new Promise((r) => setTimeout(r, 200));
+			await new Promise((r) => setTimeout(r, 0));
 
 			return function ImportedComponent() {
-				return (
-					<ThemeContext.Provider>
-						<div>2</div>
-					</ThemeContext.Provider>
-				);
+				return <div>2</div>;
 			};
 		});
 
-		const LoadableTheme = ({}) => (
+		const LoadableComponent = ({}) => (
 			<Suspense fallback={'...loading'}>
 				<LazyComponent />
 			</Suspense>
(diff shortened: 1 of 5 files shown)


test/compat/async.test.jsx
```diff
@@ -175,4 +175,46 @@ describe('Async renderToString', () => {
 		const rendered = await promise;
 		expect(rendered).to.equal('<p>ok</p>');
 	});
+
+	it('should render JSX after a urql component', async () => {
```
JoviDeCroock, line 179: We might also want to try a test where it's triple-y nested as it might warrant a loop similar to the `Array.isArray` one.

test/compat/async.test.jsx
```diff
@@ -26,10 +26,14 @@ describe('Async renderToString', () => {
 	});
 
 	it('should render JSX with nested suspended components', async () => {
-		const { Suspender: SuspenderOne, suspended: suspendedOne } =
-			createSuspender();
-		const { Suspender: SuspenderTwo, suspended: suspendedTwo } =
-			createSuspender();
+		const {
```
ccouzens, line 29: The git commit formatter hook makes these changes. Let me know if you want me to keep them as is.

test/compat/async.test.jsx
```diff
@@ -175,4 +189,41 @@ describe('Async renderToString', () => {
 		const rendered = await promise;
 		expect(rendered).to.equal('<p>ok</p>');
 	});
+
+	it('should work with setup comparable to URQL containing a lazy loaded component', async () => {
```
JoviDeCroock, line 193:
```suggestion
	it('should work with an in-render suspension', async () => {
```

JoviDeCroock (approved):

JoviDeCroock:
Have you tried breaking it by increasing the nesting even further? Just wondering 😅

test/compat/async.test.jsx
ccouzens: What do you think this would look like?

If I do something like:

```javascript
		const Context = createContext();

		let c = 0;

		const Fetcher = ({ children }) => {
			c++;
			if (c === 1) {
				throw Promise.resolve();
			}
			return <Fragment>{children}</Fragment>;
		};

		const LazyComponent = lazy(
			async () =>
				function ImportedComponent() {
					return <div>2</div>;
				}
		);

		const NestedLazyComponent = lazy(
			async () =>
				function ImportedComponent() {
					return (
						<Context.Provider>
							<Fetcher>
								<LoadableComponent />
							</Fetcher>
						</Context.Provider>
					);
				}
		);

		const LoadableComponent = ({}) => (
			<Suspense fallback={'...loading'}>
				<LazyComponent />
			</Suspense>
		);

		const NestedLoadableComponent = ({}) => (
			<Suspense fallback={'...loading'}>
				<NestedLazyComponent />
			</Suspense>
		);

		const rendered = await renderToStringAsync(
			<Context.Provider>
				<Fetcher>
					<NestedLoadableComponent />
				</Fetcher>
			</Context.Provider>
		);

		expect(rendered).to.equal(`<div>2</div>`);
```

Then the current solution continues to work

PR #383, opened 2024-07-27
chore: general performance optimisations

JoviDeCroock:
- use an `EMPTY_STR` for coercion to reduce allocation of empty strings as well as for empty string checks
- only call `encodeEntities` when the entity in question is a string, this removes always calling the function _and_ an unnecessary coercion
- ensure we leverage `x = x + y` rather than `x += y`
- ensure we leverage our `isArray` helper
- guard for some code-paths that are only possible with i.e. a class-component (avoids expensive object checks)
- skip function type properties

Results from CI (locally saw a bit larger increases, maybe node version?)

```
Text:
baseline x 259.905 ops/sec ±2.52 (83 runs sampled)
current x 273.589 ops/sec ±2.15 (87 runs sampled)
Fastest is current

SearchResults:
baseline x 1,330.724 ops/sec ±3.20 (87 runs sampled)
current x 1,365.269 ops/sec ±2.88 (89 runs sampled)
Fastest is current,baseline

ColorPicker:
baseline x 6,429.806 ops/sec ±3.09 (91 runs sampled)
current x 7,311.005 ops/sec ±3.29 (86 runs sampled)
Fastest is current

Stack Depth:
baseline x 183.456 ops/sec ±4.74 (67 runs sampled)
current x 194.845 ops/sec ±3.77 (72 runs sampled)
Fastest is current,baseline
```

### DIFF
--- a/.changeset/calm-cherries-lay.md
+++ b/.changeset/calm-cherries-lay.md
@@ -1,0 +1,5 @@
+---
+"preact-render-to-string": patch
+---
+
+General performance optimisations
--- a/benchmarks/index.js
+++ b/benchmarks/index.js
@@ -4,7 +4,7 @@ import renderToStringBaseline from 'baseline-rts';
 // import renderToString from '../src/index';
 import renderToString from '../dist/index.module.js';
 import TextApp from './text';
-// import StackApp from './stack';
+import StackApp from './stack';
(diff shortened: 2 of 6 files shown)


marvinhagemeister (approved):
Nice, this is great!

Issue #387, opened 2024-08-15
`spellcheck={false}` not working


martpet:
When using `spellcheck={false}` there is no result, but `spellcheck="false"` works (with a typescript error).

PR #388, opened 2024-08-15
fix: wrong spellcheck attribute value

marvinhagemeister:
Fixes https://github.com/preactjs/preact-render-to-string/issues/387

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -635,6 +635,9 @@ function _renderToString(
 					}
 				} else if (HTML_LOWER_CASE.test(name)) {
 					name = name.toLowerCase();
+					if (name === 'spellcheck') {
+						v = '' + v;
+					}
 				}
 			}
 		}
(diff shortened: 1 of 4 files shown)


JoviDeCroock (approved):

src/index.js
```diff
@@ -635,6 +635,9 @@ function _renderToString(
 					}
 				} else if (HTML_LOWER_CASE.test(name)) {
 					name = name.toLowerCase();
+					if (name === 'spellcheck') {
```
rschristian, line 638: Not that it necessarily needs to be done here, but I think there's at least a handful of attrs/props that could need (depending on use circumstances) to be preserved as falsy.

Might be something we need to add another regex constant for.

rschristian (approved):

src/index.js
marvinhagemeister: Yeah, good point. Let's keep out an eye for that.

PR #393, opened 2024-09-14
fix: `spellcheck` prop

rschristian:
Supports `spellcheck={false}` & removes support for `spellCheck` as we don't support it in Preact (see: https://github.com/preactjs/preact/issues/3399)

Related: #388, https://github.com/denoland/fresh/issues/2649, https://github.com/preactjs/preact/pull/4497

This adds a `HTML_ENUMERATED` rgx to stuff enumerated true/false attributes into in the future, as I brought up in #388.

### DIFF
--- a/src/lib/util.js
+++ b/src/lib/util.js
@@ -1,8 +1,9 @@
 export const VOID_ELEMENTS = /^(?:area|base|br|col|embed|hr|img|input|link|meta|param|source|track|wbr)$/;
 export const UNSAFE_NAME = /[\s\n\\/='"\0<>]/;
 export const NAMESPACE_REPLACE_REGEX = /^(xlink|xmlns|xml)([A-Z])/;
-export const HTML_LOWER_CASE = /^accessK|^auto[A-Z]|^cell|^ch|^col|cont|cross|dateT|encT|form[A-Z]|frame|hrefL|inputM|maxL|minL|noV|playsI|popoverT|readO|rowS|spellC|src[A-Z]|tabI|useM|item[A-Z]/;
-export const SVG_CAMEL_CASE = /^ac|^ali|arabic|basel|cap|clipPath$|clipRule$|color|dominant|enable|fill|flood|font|glyph[^R]|horiz|image|letter|lighting|marker[^WUH]|overline|panose|pointe|paint|rendering|shape|stop|strikethrough|stroke|spel|text[^L]|transform|underline|unicode|units|^v[^i]|^w|^xH/;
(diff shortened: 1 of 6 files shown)


rschristian:
Oh what the heck happened to the lockfile, will fix that later. Gotta get some food

src/lib/util.js
```diff
@@ -1,8 +1,9 @@
 export const VOID_ELEMENTS = /^(?:area|base|br|col|embed|hr|img|input|link|meta|param|source|track|wbr)$/;
 export const UNSAFE_NAME = /[\s\n\\/='"\0<>]/;
 export const NAMESPACE_REPLACE_REGEX = /^(xlink|xmlns|xml)([A-Z])/;
-export const HTML_LOWER_CASE = /^accessK|^auto[A-Z]|^cell|^ch|^col|cont|cross|dateT|encT|form[A-Z]|frame|hrefL|inputM|maxL|minL|noV|playsI|popoverT|readO|rowS|spellC|src[A-Z]|tabI|useM|item[A-Z]/;
-export const SVG_CAMEL_CASE = /^ac|^ali|arabic|basel|cap|clipPath$|clipRule$|color|dominant|enable|fill|flood|font|glyph[^R]|horiz|image|letter|lighting|marker[^WUH]|overline|panose|pointe|paint|rendering|shape|stop|strikethrough|stroke|spel|text[^L]|transform|underline|unicode|units|^v[^i]|^w|^xH/;
+export const HTML_LOWER_CASE = /^accessK|^auto[A-Z]|^cell|^ch|^col|cont|cross|dateT|encT|form[A-Z]|frame|hrefL|inputM|maxL|minL|noV|playsI|popoverT|readO|rowS|src[A-Z]|tabI|useM|item[A-Z]/;
+export const HTML_ENUMERATED = /^dra|spel/;
```
JoviDeCroock, line 5: Can we use a set instead, performance wise this could help a lot

JoviDeCroock (approved):

src/lib/util.js
rschristian: Sure, can do.

rschristian: Out of curiosity, what makes a set preferable here and not for the others? We certainly do make use of the patterns in a couple places, but that should just be byte saving, no?

There are certainly are a few more boolean props -> enumerated attrs, if it's the size of 2 that is what immediately concerns you, though I don't know quite how many there are. Will have to go hunting.

JoviDeCroock: I have undergone the efforts to port self-closing to a set but for the others I just don't know the amount of cases 😅

rschristian: Gotcha, thanks!

developit:
looks like this may have regressed on the SearchResults benchmark?

The fastest way to do this type of check would be an object with keys:

```js
const HTML_ENUMERATED = {
  draggable: true,
  spellcheck: true
};

// <snip>

} else if (
	(name[4] === '-' || name in HTML_ENUMERATED) &&
	v != null
) {
```

PR #398, opened 2024-09-22
refactor: Regex to sets

rschristian:
Might be marginally faster? Probably easier to maintain at the very least -- easier to tell what properties are being replaced and which aren't.

I did this by comparing against Preact's types, mostly, with a few corrections needed for the SVG types.


src/index.js
```diff
@@ -5,7 +5,7 @@ import {
 	NAMESPACE_REPLACE_REGEX,
 	HTML_LOWER_CASE,
 	HTML_ENUMERATED,
-	SVG_CAMEL_CASE,
+	SVG_KEBAB_CASE,
```
rschristian, line 8: Renamed for consistency.

`HTML_LOWER_CASE` is converting camel-cased names to lower-cased, therefore `SVG_CAMEL_CASE` stands out as it's converting camel-cased names to kebab-cased.

Issue #406, opened 2025-01-08
useId stuck in infinite loop with streaming render


f0x52:
- [x] Check if updating to the latest Preact version resolves the issue

**Describe the bug**
I encountered an infinite loop in the `useId` hook, when trying to debug a different issue relating to useId/Suspense.  

```jsx
<div>
  <Suspense fallback="suspense-1">
    <ComponentOne/>
  </Suspense>
  <Suspense fallback="suspense-1">
    <ComponentTwo/>
  </Suspense>
</div>
```
With this component tree, where `ComponentOne` and `ComponentTwo` both load a `lazy()` component, as soon as rendering resumes on the first Suspense boundary, the `useId` hook inside `ComponentOne` loops infinitely.

**To Reproduce**
https://codesandbox.io/p/devbox/preact-useid-infinite-loop-4clmdc?workspaceId=ws_Hyb2mUcrQ6TVf2LBce8QH7

Launching the run/debug configuration 'Run program' will try to do a server-side streaming render with two Suspense boundaries, both of which have a component that logs the output of the `useId` hook, and then loads a `lazy()` wrapped component.

As debug output you will see:
```
render ComponentOne
  ComponentOne useId: P0-0
render ComponentTwo
  ComponentTwo useId: P0-1
resolved async component
render ComponentOne
```
After which the process freezes (and Codesandbox pops up a 'CPU usage 100%' in the top bar). If you pause the debug, you can see it's stuck in an infinite while loop in useHook, because `root._parent` is a circular reference.
https://github.com/preactjs/preact/blob/1226aae23826e0d703b520cb764749e04f571043/hooks/src/index.js#L431-L433

**Expected behavior**
Render finished, expected output
```
render ComponentOne
  ComponentOne useId: P0-0
render ComponentTwo
  ComponentTwo useId: P0-1
resolved async component
render ComponentOne
  ComponentOne useId: P0-0
resolved async component
render ComponentTwo
  ComponentTwo useId: P0-1
```

f0x52:
Think I chased it down to the following code:
the `renderer.onError` callback `renderChild` argument sets the vnode that whose render the promise was caught in as the parent (5th arg to _renderToString)
https://github.com/preactjs/preact-render-to-string/blob/2da38a7ac4bccbe00865265c152bccc4c8098cea/src/index.js#L483-L493
The error handler walks up to find the Suspense boundary `vnode`
[...]

PR #408, opened 2025-01-08
Use suspense boundary vnode as parent for subtree re-render

f0x52:
Fixes https://github.com/preactjs/preact-render-to-string/issues/406

### DIFF
--- a/test/compat/render-chunked.test.js
+++ b/test/compat/render-chunked.test.js
@@ -6,7 +6,6 @@ import { renderToChunks } from '../../src/lib/chunked';
 import { createSubtree, createInitScript } from '../../src/lib/client';
 import { createSuspender } from '../utils';
 import { VNODE, PARENT } from '../../src/lib/constants';
-import { Deferred } from '../../src/lib/util';
 
 describe('renderToChunks', () => {
 	it('should render non-suspended JSX in one go', async () => {
--- a/test/compat/render-chunked.test.js
+++ b/test/compat/render-chunked.test.js
@@ -1,10 +1,12 @@
 import { h } from 'preact';
 import { expect } from 'chai';
 import { Suspense } from 'preact/compat';
+import { useId } from 'preact/hooks';
 import { renderToChunks } from '../../src/lib/chunked';
 import { createSubtree, createInitScript } from '../../src/lib/client';
 import { createSuspender } from '../utils';
 import { VNODE, PARENT } from '../../src/lib/constants';
+import { Deferred } from '../../src/lib/util';
 
 describe('renderToChunks', () => {
 	it('should render non-suspended JSX in one go', async () => {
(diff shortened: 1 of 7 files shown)


JoviDeCroock (approved):
This does look good, we'll need a changeset and a few tests to prevent regressions. You might need to update the types of the renderer as well

f0x52:
Would this be a `patch` changeset?

JoviDeCroock:
You should do the useId test and leverage the abortController to abort after i.e 1 second and fail

f0x52:
I don't think that's possible either? The synchronous while loop inside the useId hook completely blocks the Node.js event loop, so as far as I know there is no way to abort the abortController while it's in the loop; timeouts/intervals won't even be called. [Tripwire](https://github.com/tjanczuk/tripwire) as mentioned in the mocha issue uses a native extension to be able to interrupt unbreakable loops, but including a native extension + build step doesn't seem ideal either.

Perhaps it would be good to add a check in while loops that could otherwise run forever on erroneous node trees; so far I found this in useId, and the _catchError handler, though that's probably a performance tradeoff

JoviDeCroock:
I mean, it works now so we should not care about a test loop.... we should write a test, i can write it if you prefer.

f0x52:
This test also expects the current behavior where `useId`'s are scoped to the rendering root/boundary, resulting in id P0-0 for the first element outside the suspense boundary, and P0-0 for the first element inside as well.

test/compat/render-chunked.test.js
```diff
@@ -66,4 +68,85 @@ describe('renderToChunks', () => {
...
+			'<div><!--preact-island:16-->loading...<!--/preact-island:16--></div>',
+			'<div hidden>',
+			createInitScript(1),
+			createSubtree('16', '<p>it works</p>'),
+			'</div>'
+		]);
+	});
+
+	it('should support using useId hooks inside a suspense boundary', async () => {
+		const { Suspender, suspended } = createSuspender();
```
JoviDeCroock, line 121: You can add children here, so if we do the following twice, then you can actually assert the behaviour of your issue where you don't want two adjacent trees to have the same id

```suggestion
		const { Suspender, suspended } = createSuspender(<ComponentWithId />);
```

f0x52: `createSuspender` doesn't take any arguments?
But changing it to
```jsx
				<ComponentWithId />
				<Suspense fallback="loading...">
					<Suspender>
						<ComponentWithId />
					</Suspender>
					<Suspender>
						<ComponentWithId />
					</Suspender>
				</Suspense>
```
could work, then the expected result is P0-0, P0-0, P0-1

JoviDeCroock: It does on main

f0x52: am I missing something here? https://github.com/preactjs/preact-render-to-string/blob/main/test/utils.jsx#L18

f0x52: If you want to finish it up, that sounds good to me!

PR #418, opened 2025-02-08
feat: Support `dangerouslySetInnerHTML` on Fragments

rschristian:
Might/probably will need work in core as well (though arguably this is more useful on the server? not sure), definitely could do with some more tests yet.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
 	"name": "preact-render-to-string",
-	"version": "6.5.12",
+	"version": "6.5.13",
 	"lockfileVersion": 3,
 	"requires": true,
 	"packages": {
 		"": {
 			"name": "preact-render-to-string",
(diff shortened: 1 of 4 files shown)


nwalters512:
> Might/probably will need work in core as well (though arguably this is more useful on the server? not sure)

For a project that's using SSR + hydration, I'd love for this to work the same during both static rendering and on the client!

rschristian:
> For a project that's using SSR + hydration, I'd love for this to work the same during both static rendering and on the client!

It should "hydrate" just fine as-is, by virtue of not actually having anything to hydrate; there's no hydrating non-Preact components and we don't wipe out the DOM on mismatches. Hydrating `<Fragment dangerouslySetInnerHTML={{ __html: '<h1>Hello World!</h1>' }}>` is basically a no-op.

Issue #422, opened 2025-08-03
Not working properly when using signals on "disabled" attribute


hronro:
```jsx
import { render } from 'preact-render-to-string';
import { signal } from '@preact/signals-core'
import { h } from 'preact';
/** @jsx h */

const disabled = signal(false);

let vdom = <input class='foo' disabled={disabled} />;

let html = render(vdom);
console.log(html);
```

The actual output is `<input class="foo" disabled="false"/>`, which is equal to `<input class="foo" disabled/>`.
The expected output should be `<input class="foo"/>`.

PR #424, opened 2025-08-06
Detect and unwrap signal properties

JoviDeCroock:
Fixes https://github.com/preactjs/preact-render-to-string/issues/422

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -8,6 +8,7 @@ import {
 	SVG_CAMEL_CASE,
 	createComponent
 } from './lib/util.js';
+import { Signal } from '@preact/signals-core';
 import { options, h, Fragment } from 'preact';
 import {
 	CHILDREN,
@@ -559,6 +560,7 @@ function _renderToString(
 
 	for (let name in props) {
 		let v = props[name];
+		v = isSignal(v) ? v.value : v;
 
 		if (typeof v == 'function' && name !== 'class' && name !== 'className') {
 			continue;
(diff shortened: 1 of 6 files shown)


src/index.js
```diff
@@ -740,3 +742,7 @@ const SELF_CLOSING = new Set([
 export default renderToString;
 export const render = renderToString;
 export const renderToStaticMarkup = renderToString;
+
+function isSignal(x) {
+	return x instanceof Signal;
+}
```
marvinhagemeister, line 748: Wondering if we should use duck-typing here like https://github.com/denoland/fresh/blob/e65643c482de9b5afee5d52af3ccc88941c35b5c/src/runtime/server/preact_hooks.tsx#L355-L362 . That way we can keep rts dependency free and don't have to keep it in sync with major `@preact/signals` version increases.

marvinhagemeister: Not sure if `x !== null && typeof x === "object"` is more or less expensive than an `instanceof` check. I let you decide on that one.

marvinhagemeister (approved):
LGTM

src/index.js
```diff
@@ -8,6 +8,7 @@ import {
 	SVG_CAMEL_CASE,
 	createComponent
 } from './lib/util.js';
+import { Signal } from '@preact/signals-core';
```
rschristian, line 11: Looks like this was accidentally left around, will need to be removed but I gotta get to bed -- I can do it tomorrow but wanted to drop a mention at least so we don't release this.

hronro:
> We'll release it when we're ready, please be patient.
> 
> You can use `patch-package` or `pnpm patch` in the meantime if you need it, should be pretty straightforward to add.

Thanks for your reply. I'm using Deno with the [Fresh](https://fresh.deno.dev) framework, which isn't easy to patch a npm dependency, neither `pactch-package` nor `pnpm patch` works for me. Anyway, I appreciate the work you guys have done and hope there could be a new release soon.

PR #430, opened 2025-09-29
Fix JSX renderer to correctly render enumerated properties

### LINKED ISSUE #429: `jsx` renderer with `jsx: false` does not render `draggable` correctly
```tsx
import { h } from 'preact';
import { render } from 'preact-render-to-string/jsx';

const node = h('div', { draggable: true });
console.log(render(node, {}, { pretty: true, jsx: false }));
```

Output:
```html
<div draggable></div>
```

According to the Preact types (https://github.com/preactjs/preact/blob/5e183563a101de2d7a750e7c606d0b13677637f1/src/dom.d.ts#L1267):

```
	draggable?: Signalish<boolean | undefined>;
```

From https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Global_attributes/draggabler:

> Warning: This attribute is [enumerated](https://developer.mozilla.org/en-US/docs/Glossary/Enumerated) and not Boolean. A value of true or false is mandatory, and shorthand like `` is forbidden. The correct usage is ``.

Either this renderer needs to be fixed, or the upstream types need to be fixed (https://github.com/preactjs/preact/issues/4926).

reteps:
Closes #429

### DIFF
--- a/src/pretty.js
+++ b/src/pretty.js
@@ -24,6 +24,7 @@ const UNNAMED = [];
 
 const EMPTY_ARR = [];
 const EMPTY_STR = '';
+
 /**
  * Render Preact JSX + Components to a pretty-printed HTML-like string.
  * @param {VNode} vnode	JSX Element / VNode to render
--- a/src/pretty.js
+++ b/src/pretty.js
@@ -9,6 +9,7 @@ import {
 	VOID_ELEMENTS,
 	NAMESPACE_REPLACE_REGEX,
 	SVG_CAMEL_CASE,
+	HTML_ENUMERATED,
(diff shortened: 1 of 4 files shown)


reteps (commented):
self-review

src/pretty.js
```diff
@@ -22,7 +23,7 @@ import { options, Fragment } from 'preact';
 const UNNAMED = [];
 
 const EMPTY_ARR = [];
-
+const EMPTY_STR = '';
```
rschristian, line 26: Nit:

```suggestion
const EMPTY_STR = '';

```

rschristian (approved):
Thanks!

Issue #438, opened 2025-12-31
useId produces duplicate IDs in JSX mode


nwalters512:
Self-contained reproduction: https://github.com/nwalters512/preact-use-id-repro

```tsx
import { useId } from "preact/compat";
import renderToString from "preact-render-to-string/jsx";

function Foo() {
  const id = useId();
  return <div id={id} />;
}

const App = () => {
  return (
    <>
      <Foo />
      <Foo />
    </>
  );
};

const html = renderToString(<App />);
console.log(html);
```

This produces the following output:

```
<div id="P0-0"></div>
<div id="P0-0"></div>
```

When using the non-JSX entrypoint, it produces unique IDs as expected:

```
<div id="P0-0"></div><div id="P0-1"></div>
```

As far as I can tell, this occurs because the JSX renderer never sets `vnode[PARENT]` like the other one does:

https://github.com/preactjs/preact-render-to-string/blob/534a37ac8e1386b10ea646b5477a1836e7582e81/src/index.js#L298

So when the `useId()` hook tries to walk to the root element, there simply is no parent:

https://github.com/preactjs/preact/blob/bf7a195ac4b1706468e876e41b27428e3d8a08f3/hooks/src/index.js#L429-L432

PR #439, opened 2025-12-31
Fix useId in JSX renderer

nwalters512:
This PR fixes https://github.com/preactjs/preact-render-to-string/issues/438. It includes two new tests, both of which fail on `main` and pass with the changes included here.

For transparency: the diff itself was largely written by Claude Code with Claude Opus 4.5. I reviewed the output for correctness and consistency with the rendering process in `src/index.js`.

### DIFF
--- a/.changeset/friendly-queens-juggle.md
+++ b/.changeset/friendly-queens-juggle.md
@@ -1,0 +1,5 @@
+---
+'preact-render-to-string': patch
+---
+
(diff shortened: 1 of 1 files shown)


rschristian:
Thanks!

Will take a better look shortly, but if you wouldn't mind, please run `npx @changesets/cli`. Changesets builds up our change log and also helps automate releases.

PR #442, opened 2026-01-08
feat!: stream Suspense chunks via declarative shadow dom

jacob-ebey:


jacob-ebey:
This doesn't work quite the way I had thought. This won't work but was a fun learning experience.

jacob-ebey:
This doesn't work with Suspense semantics, but it did lead to a potential new primitive (Shadow) where it's non-template children render in the order they land allowing OOO streaming semantics within a single view. Deterministic UI order can be achieved with slot names even on `<Component slot="target">`.

```tsx
function Wrapper({ children }: { children?: JSX.Children }) {
  return (
    <Shadow ext="section">
      <Shadow.Template>
        <p>before</p>
        <slot>loading</slot>
        <p>after</p>
      </Shadow.Template>
      {children}
    </Shadow>
  );
}

async function Component({
  children,
  delay,
}: {
  children?: JSX.Children;
  delay: number;
}) {
  await new Promise((r) => setTimeout(r, delay * 500));
  return <p>resolved {children}</p>;
}
```

```jsx
<html>
  <head></head>
  <body>
    <h1>Hello, World!</h1>
    <Wrapper>
      <Component delay={3}>1</Component>
      <Component delay={4}>2</Component>
      <Wrapper>
        <Component delay={2}>1.1</Component>
        <Component delay={1}>2.1</Component>
      </Wrapper>
    </Wrapper>
  </body>
</html>
```

PR #464, opened 2026-06-09
fix: call root hook during server render

JoviDeCroock:
## Summary

- call Preact's mangled `options._root` hook from `renderToString()` and `renderToStringAsync()` root renders
- pass a DOM-like fake parent so debug's root validation accepts the server-render parent shape
- add a focused regression for root hook invocation

The Suspense/useId mask allocation change does not belong in this repo; that should live in the Preact PR.

> [!NOTE]
> Co-Authored by gpt5.5.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -21,9 +21,12 @@ import {
 	NEXT_STATE,
 	PARENT,
 	RENDER,
+	ROOT,
 	SKIP_EFFECTS,
 	VNODE,
-	CATCH_ERROR
+	CATCH_ERROR,
+	CHILD_DID_SUSPEND,
+	MASK
 } from './lib/constants.js';
 
 const EMPTY_OBJ = {};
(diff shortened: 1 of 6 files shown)


rschristian (approved):

src/index.js
```diff
@@ -80,6 +85,7 @@ export function renderToString(vnode, context, _rendererState) {
 
 	const parent = h(Fragment, null);
 	parent[CHILDREN] = [vnode];
+	callRootHook(vnode, { _children: null, nodeType: 1 });
```
marvinhagemeister: Nit: should we properly connect the tree here like:
```js
{ [CHILDREN]: parent, nodeType: 1 }
```
Otherwise the second arg is just a dummy value.

marvinhagemeister: also wondering if we should just inline `callRootHook` since it's so small. Not sure if it will make a difference in v8 though.

marvinhagemeister (approved):

PR #469, opened 2026-07-11
Perf tuning

JoviDeCroock:
## Summary

This tunes several hot paths in synchronous, asynchronous, JSX, and streaming rendering:

- avoid a redundant await for synchronous results returned through `renderToStringAsync`
- replace entity and attribute regular-expression work with targeted string and prefix scans
- reuse component prototype lookups and serialize numeric children directly
- specialize primitive array children and skip function-valued children without recursive renderer calls
- parse XML, XMLNS, and XLINK attribute prefixes without regex replacement or prefix slicing
- replace self-closing element Set lookups with a direct switch
- scan and flatten asynchronous render results directly and wrap suspense arrays in one pass
- remove quadratic suspension cleanup and skip abort-race allocation when no abort signal is present
- correct the async benchmark so Benchmark.js waits for each render to settle

## Measured impact

- small synchronous trees through `renderToStringAsync`: 16.7–19.8% faster
- plain text serialization: 18.4–21.3% faster
- short safe strings: 7.7–15.8% faster; short escaped strings: 10.6–17.4% faster
- HTML attribute-heavy rendering: 8.0–8.1% faster; SVG attributes: 3.5–4.9% faster
- deep function-component trees: 6.1–6.8% faster; class-component trees: 2.3–5.9% faster
- namespace-heavy SVG rendering: 2.39–2.40× throughput
- namespace-heavy JSX rendering: 2.04–2.22× throughput
- numeric child lists: 9.1–14.2% faster
- suspension cleanup: 10.3–12.8% faster at 5,000 concurrent boundaries
- non-abortable streams: 1.8–3.4% faster with one suspended boundary and 2.6–3.2% faster with 10–100 boundaries

## Additional D8 measurements

Measured against the previous PR head `b8d7377` using seven alternating fresh-process runs per side and isolated hot-path filters:

- primitive child arrays: 18.7% faster internally and 10.6% faster through `renderToString`
- arrays containing function values: 39.5% faster internally and 14.9% faster through `renderToString`
- empty strings: 2.6% faster; ordinary text: 6.8% faster
- XML, XMLNS, and XLINK prefix checks: 51.9–74.5% faster; SVG namespace rendering: 4.7% faster
- promise detection in async result arrays: 9.5–11.1% faster
- one-level async result flattening: 59.7% faster
- suspense array wrapping: 58.0% faster
- void-element rendering: 2.2% faster
- four concurrently suspending siblings: 0.75% faster

### DIFF
--- a/src/lib/util.js
+++ b/src/lib/util.js
@@ -40,16 +40,22 @@ export function isDirty(component) {
 /** @param {string} str */
 export function encodeEntities(str) {
 	// Skip all work for strings with no entities needing encoding:
-	if (
-		str.length === 0 ||
-		(str.indexOf('"') === -1 &&
-			str.indexOf('&') === -1 &&
-			str.indexOf('<') === -1)
-	)
+	let i = 0;
+	if (str.length < 8) {
+		for (; i < str.length; i++) {
+			const char = str.charCodeAt(i);
+			if (char === 34 || char === 38 || char === 60) break;
+		}
+		if (i === str.length) return str;
+	} else if (
+		str.indexOf('"') === -1 &&
+		str.indexOf('&') === -1 &&
+		str.indexOf('<') === -1
+	) {
 		return str;
+	}
 
 	let last = 0,
-		i = 0,
 		out = '',
 		ch = '';
 
--- a/src/lib/util.js
+++ b/src/lib/util.js
@@ -37,13 +37,16 @@ export function isDirty(component) {
 	return component[DIRTY] === true;
 }
 
-// DOM properties that should NOT have "px" added when numeric
-const ENCODED_ENTITIES = /["&<]/;
-
 /** @param {string} str */
 export function encodeEntities(str) {
 	// Skip all work for strings with no entities needing encoding:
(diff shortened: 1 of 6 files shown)


src/lib/util.js
```diff
@@ -37,16 +37,25 @@ export function isDirty(component) {
...
 
-// DOM properties that should NOT have "px" added when numeric
-const ENCODED_ENTITIES = /["&<]/;
-
 /** @param {string} str */
 export function encodeEntities(str) {
 	// Skip all work for strings with no entities needing encoding:
-	if (str.length === 0 || ENCODED_ENTITIES.test(str) === false) return str;
+	let i = 0;
+	if (str.length < 8) {
```
developit, line 44: Why 8?

JoviDeCroock: I tried all numbers and larger than 8 seemed to be slower in V8 :joy:

> Use a character-code loop for strings shorter than eight characters and retain native indexOf scans for longer strings. Escaped short strings continue encoding from the first entity found by the fast-path scan.

PR #472, opened 2026-08-24
feat: async nonce support

jacob-ebey:
Can't use a CSP headers with async rendering at the moment without allowing `unsafe-inline`.

### DIFF
--- a/src/stream.js
+++ b/src/stream.js
@@ -5,10 +5,11 @@ import { renderToChunks } from './lib/chunked.js';
 
 /**
  * @param {import('preact').VNode} vnode
+ * @param {any} [options]
  * @param {any} [context]
  * @returns {RenderStream}
  */
-export function renderToReadableStream(vnode, context) {
+export function renderToReadableStream(vnode, options, context) {
 	/** @type {Deferred<void>} */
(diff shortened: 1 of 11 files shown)


src/stream.js
```diff
@@ -5,10 +5,11 @@ import { renderToChunks } from './lib/chunked.js';
 
 /**
  * @param {import('preact').VNode} vnode
+ * @param {any} [options]
  * @param {any} [context]
  * @returns {RenderStream}
  */
-export function renderToReadableStream(vnode, context) {
+export function renderToReadableStream(vnode, options, context) {
```
JoviDeCroock, line 12: This qualifies as a breaking change, not sure whether we can do that atm 😅

jacob-ebey: can we roll it into context?

JoviDeCroock: Maybe we can get away with your current change if we strictly type `options` to dis-allow any other properties as we have considered this experimental and usage is pretty low.

jacob-ebey: I'll leave it up to you. Feel free to edit this PR if you'd like or create a new one.

JoviDeCroock (approved):

